PR #135, opened 2017-06-26
Removed default args to derivation

dmjio:


dmjio:
This will break the build until we get caching for ghcjs8 packages

PR #150, opened 2017-07-04
Added release.nix for hydra, added hydra badge

dmjio:
🍜


dmjio:
Travis has been remove as well.

PR #185, opened 2017-07-14
Add nix gen'd from stack files

dmjio:
This will test that the stack files can be built and that `miso` tests can run against them.
- [x] add nix gen'd from stack files
- [x] import and override in `default.nix`
- [x] include in build, and tests

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -4,7 +4,13 @@ let
   inherit (nixpkgs.haskell.packages) ghc802 ghcjs;
   inherit (nixpkgs.lib) overrideDerivation optionalString;
   inherit (nixpkgs.stdenv) isDarwin;
-  inherit (nixpkgs) phantomjs2 closurecompiler;
+  inherit (nixpkgs) phantomjs2 closurecompiler writeScriptBin fetchFromGitHub;
+  nixFromStack = import (fetchFromGitHub {
+    owner = "dmjio";
+    repo = "nixFromStack";
+    rev = "e6edf8bfc4dd37398637c2f4b9a39d5d2784e9df";
+    sha256 = "17v2r8mafk30c3djlh04vvg954x3wq3snk4dc54wb9gi8acqhxsk";
+  }) { pkgs = nixpkgs; };
   miso-ghc = ghc802.callPackage ./miso-ghc.nix { };
   miso-ghcjs = (ghcjs.callPackage ./miso-ghcjs.nix { }).overrideDerivation (drv: {
     doCheck = tests && !isDarwin;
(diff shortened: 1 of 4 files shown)


default.nix
```diff
@@ -4,7 +4,13 @@ let
   inherit (nixpkgs.haskell.packages) ghc802 ghcjs;
   inherit (nixpkgs.lib) overrideDerivation optionalString;
   inherit (nixpkgs.stdenv) isDarwin;
-  inherit (nixpkgs) phantomjs2 closurecompiler;
+  inherit (nixpkgs) phantomjs2 closurecompiler writeScriptBin fetchFromGitHub;
+  nixFromStack = import (fetchFromGitHub {
```
adnelson: Importing a derivation is problematic; it requires nix to build (and/or fetch) objects just to *evaluate* the expression, not even to build it. For example, create a file `~/test.nix`, with

```nix
let
  nixpkgs = import <nixpkgs> {};
in

import (nixpkgs.fetchFromGitHub {
    owner = "dmjio";
    repo = "nixFromStack";
    rev = "e6edf8bfc4dd37398637c2f4b9a39d5d2784e9df";
    sha256 = "17v2r8mafk30c3djlh04vvg954x3wq3snk4dc54wb9gi8acqhxsk";
}) { pkgs = nixpkgs; }
```

Then run `nix-build ~/test.nix --dry-run`

Even though it's a dry-run build, you will still see nix fetching and/or building objects (unless it's already been fetched before) because it can't compute the hash of the above expression without doing so. Since in general people expect instantiation to be pure and quick, and because of the potential for confusing errors, in general (or entirely?) this pattern is avoided at least in `nixpkgs`. 

There are a few alternatives which I think are a bit better:

1. Use a submodule to the `nixFromStack` repo in the `miso` repo.

```bash
$ git submodule add https://github.com/dmjio/nixFromStack
$ cd nixFromStack && git checkout e6edf8bfc4dd37398637c2f4b9a39d5d2784e9df
```

Then this section can simply be changed to `nixFromStack = import ./nixFromStack {pkgs = nixpkgs;};`

2. Reference the commit of `nixFromStack` as the `src` of another derivation, and instead define `nixFromStack` using `mkDerivation`.

I'd suggest the first one since submodules are a pretty common way to encode deterministic linkage between different repos.

Also the same approach could be used to define the `stackage-packages` files defined below, so that they wouldn't need to be copied into this repo (which seems like the wrong place?)

adnelson: Of course in practice this will probably be a non-issue, so there's that.

dmjio:
Postponing this

PR #205, opened 2017-07-27
Add quick start section

dmjio:


README.md
```diff
@@ -116,6 +134,220 @@ viewModel x = div_ [] [
...
+extra-deps: []
+
+setup-info:
+  ghcjs:
+    source:
+      ghcjs-0.2.0.9006020_ghc-7.10.3:
+         url: http://ghcjs.tolysz.org/lts-6.20-9006020.tar.gz
+         sha1: a6cea90cd8121eee3afb201183c6e9bd6bacd94a
+
+- location:
```
cocreature: Is there a reason to choose this specific commit? If not I would suggest to just add an entry in the `extra-deps` section that fetches `miso-0.4.0.0` from hackage.

dmjio: @cocreature I was wondering the same, but wasn't sure how to specify to fetch from `hackage`.

dmjio: 🤦‍♂️ , can just add `miso-0.4.0.0` to `extra-deps`

dmjio:
@cocreature 🍻

PR #208, opened 2017-07-27
Add clearBody, fixes #206

### LINKED ISSUE #206: Create clearBody
During `GHCJSi` sessions, the body can be redrawn twice. To stop this, we should expose the following function, and let the user run it in the REPL.

```haskell
foreign import javascript unsafe "document.body.innerHTML = ''"
  clearBody :: IO ()
```

cocreature:


dmjio:
Nice! we should also probably export this as well. Maybe we should make a `Miso.Dev` module in `ghcjs-src`, and then re-export it as well. 

```haskell
module Miso.Dev (clearBody) where

import Miso.FFI (clearBody)
```

and then re-export `Miso.Dev` from `Miso.hs` in `ghcjs-src`.

dmjio:
@cocreature beautiful, cheers! 🍻

PR #210, opened 2017-07-27
Add sse example, fixes #196

### LINKED ISSUE #196: Add sse example
Using as source URL http://sapid.sourceforge.net/ssetest/webkit.events.php

cocreature:
I’m opening this not because it’s finished (although it does work) but so that we have something to look at while we flesh out the details.

1. The examples is deliberately simple and dumb. Personally I like that way since it makes it very clear what this example is focusing on but I can try coming up with something a bit more complex if you prefer that.
2. Currently the backend is just a separate application. Should I try to integrate it in the `haskell-miso.org` backend? The URl and the port also need to be updated before this is merged (I just needed something to test it :))

### DIFF
--- a/examples/sse/client.nix
+++ b/examples/sse/client.nix
@@ -1,0 +1,16 @@
+{ mkDerivation, aeson, base, containers, miso, network-uri, servant
+, stdenv
+}:
+mkDerivation {
+  pname = "sse";
+  version = "0.1.0.0";
+  src = ./.;
+  isLibrary = false;
+  isExecutable = true;
+  executableHaskellDepends = [
+    aeson base containers miso network-uri servant
+  ];
+  homepage = "https://sse.haskell-miso.org";
+  description = "Server sent events example";
+  license = stdenv.lib.licenses.bsd3;
+}
--- a/examples/sse/sse.cabal
+++ b/examples/sse/sse.cabal
@@ -1,0 +1,60 @@
+name:                sse
+version:             0.1.0.0
+synopsis:            Server sent events example
+homepage:            https://sse.haskell-miso.org
+license:             BSD3
+license-file:        LICENSE
+author:              David Johnson
+maintainer:          djohnson.m@gmail.com
+copyright:           David Johnson (c) 2017
+category:            Web
+build-type:          Simple
+cabal-version:       >=1.10
+
+executable server
+  main-is:
+    Main.hs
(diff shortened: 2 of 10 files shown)


dmjio:
@cocreature you're pretty awesome right now.

For the `sse` example, might be best to give it it’s own cabal file (similar to how `examples/haskell-miso.org` does it) in `examples/sse`, then we can copy over the NixOS `config.nix` and `module.nix` scripts from `examples/haskell-miso.org` (update them appropriately i.e `https://sse.haskell-miso.org`) and I’ll host it from my apt.

dmjio:
So, after moving the `sse` executable sections to their own cabal file (`examples/sse/sse.cabal`), I'd just call `cabal2nix --compiler ghcjs . > client.nix` and `cabal2nix . > server.nix`, then just copy exactly what's in `examples/haskell-miso.org/default.nix`, and then call `nix-build`. It should "just work"

cocreature:
Alright this is now a separate package including the nix setup. I’ve also switched to using proper routes and made it an “isomorphic” app.

examples/sse/client.nix
```diff
@@ -0,0 +1,16 @@
...
+mkDerivation {
+  pname = "sse";
+  version = "0.1.0.0";
+  src = ./.;
+  isLibrary = false;
+  isExecutable = true;
+  executableHaskellDepends = [
+    aeson base containers miso network-uri servant
+  ];
+  homepage = "https://haskell-miso.org";
```
dmjio: can we make this `https://sse.haskell-miso.org`?

examples/sse/sse.cabal
```diff
@@ -0,0 +1,60 @@
+name:                sse
+version:             0.1.0.0
+synopsis:            Server sent events example
+homepage:            https://haskell-miso.org
```
dmjio: https://haskell-miso.org -> https://sse.haskell-miso.org

dmjio:
@cocreature nice work, added a few comments, and can we also copy over `config.nix` and `module.nix`, and make sure we move `https://haskell-miso.org` -> `https://sse.haskell-miso.org`

cocreature:
I’ve added `config.nix` and `module.nix` and changed the URL but I don’t have a way to test these files atm so hopefully I didn’t screw up too badly ;)

PR #215, opened 2017-07-29
Init WebGL

dmjio:


examples/webgl/webgl.cabal
```diff
@@ -0,0 +1,31 @@
+name:                webgl
+version:             0.1.0.0
+synopsis:            https://webgl.haskell-miso.org
+description:         WebGL example for the Miso web framework, based on three.js
+homepage:            https://webgl.haskell-miso.org
+license:             BSD3
+license-file:        LICENSE
```
cocreature: This file seems to be missing which breaks the build of this example.

dmjio:
Closing in favor of current three.js implementation.

PR #217, opened 2017-07-30
Improve documentation of `Effect`

cocreature:
I remember being slightly confused about whether actions block so it
probably makes sense to point out in the docs that they don’t block.


dmjio:
Sounds great. Maybe add `All effects are processed asynchronously`.

cocreature:
> All effects are processed asynchronously.

Isn’t that a bit redundant given that I already added `The action is always run in a new thread so there is no risk of accidentally blocking the application.`?

dmjio:
Yea, I suppose it's redundant. Ok I'll forego the bikeshedding :)

PR #224, opened 2017-08-01
Added ExtendedDefaultRules & OverloadedStrings to sample code

saurabhnanda:


cocreature:
The example seems to compile just fine without those extensions enabled. What error are you seeing?

saurabhnanda:
```
/Users/saurabhnanda/projects/js-framework-benchmark/miso-ghcjs-keyed/app/Main.hs:97:86: error:
    • Ambiguous type variable ‘str1’ arising from a use of ‘text’
      prevents the constraint ‘(Miso.String.ToMisoString
                                  str1)’ from being solved.
      Probable fix: use a type annotation to specify what ‘str1’ should be.
      These potential instances exist:
        instance Miso.String.ToMisoString String
          -- Defined in ‘Miso.String’
        ...plus six instances involving out-of-scope types
        (use -fprint-potential-instances to see them all)
    • In the expression: text "Create 1,000 rows"
      In the second argument of ‘button_’, namely
        ‘[text "Create 1,000 rows"]’
      In the expression:
        button_
          [type_ "button", class_ "btn btn-primary btn-block", id_ "run"]
          [text "Create 1,000 rows"]
```

Could it depend on other factors like the stack/cabal file of the project?

cocreature:
The error message seems to suggest that you are not compiling the sample code that you are modifying here but something else? Ofc different code can require different extensions but I don’t think that’s a good reason to modify the sample. If you have been confused by that error message then maybe the haddocks for `ToMisoString` should be modified.

saurabhnanda:
@cocreature It's not the sample code that I was compiling per-se. But I'm assuming that others, like me, would be using that as a starting point and would jump into different directions from there. 

Should we add it at both places to reduce the chances of this error cropping up? String-related issues should be last thing that a new dev should trip-up on!

cocreature:
I don’t think it should be added as *code* to the example. Adding a note that it might be necessary in more complex examples seems fine but it should be clear that it’s not necessary for that specific example.

dmjio:
Yea, I don't recommend this extension actually. It does solve a pain point when using overloaded strings and wanting to use `JSString` by default. But it introduces copious warnings. It might be better to use `import Miso.String (pack)`. On the server `MisoString ~ Text`, on the client `MisoString ~ JSString`

dmjio:
Closing, in favor of using `pack`

saurabhnanda:
> Yea, I don't recommend this extension actually. It does solve a pain point when using overloaded strings and wanting to use JSString by default. But it introduces copious warnings. It might be better to use import Miso.String (pack). On the server MisoString ~ Text, on the client MisoString ~ JSString

> Closing, in favor of using pack

I just checked the [TodoMVC sample](https://github.com/dmjio/miso/blob/da9b2a82c3ef1ee81ba5e16f27d9932b9626e631/examples/todo-mvc/Main.hs) and it uses these extension. Writing code without these extensions would force one to use `pack` on every string literal, which is pretty irritating.

Is there any way to turn off the warning with a compiler pragma?

dmjio:
There's a performance issue with `GHCJS` related to overloaded strings that should be resolved first. I recommend just using `pack`. The todo-mvc example should be migrated to it. 

https://github.com/ghcjs/ghcjs/issues/599

> Is there any way to turn off the warning with a compiler pragma?

I'd recommend checking the ghc user guide.

saurabhnanda:
> I recommend just using pack.

I'm still stuck here. If I remove the `OverloadingStrings` extension, *everything* starts complaining, eg:

```
/Users/saurabhnanda/projects/js-framework-benchmark/miso-ghcjs-keyed/app/Main.hs:170:28: error:
    • Couldn't match type ‘[Char]’ with ‘JSString’
      Expected type: MisoString
        Actual type: [Char]
    • In the first argument of ‘type_’, namely ‘"button"’
      In the expression: type_ "button"
      In the first argument of ‘button_’, namely
        ‘[type_ "button", class_ "btn btn-primary btn-block",
          id_ "runlots", onClick (CreateRows 10000)]’
```

This would force one to litter `pack` all around the HTML generation code!

Is there any other way around this?

dmjio:
@saurabhnanda, yes, but it's a large change. So bear with me. But it gets around `pack` everywhere.

PR #227, opened 2017-08-01
Support composable update functions

cocreature:
This is an implementation of the ideas that I was trying to explain to you yesterday. I tried splitting the individual changes up into small self-contained commits so hopefully it’s easy to digest but I’ll give a quick summary anyway:
* `Effect` has been generalized to allow an arbitrary number of actions instead of just `0` and `1`. This is a breaking change but the smart constructors are unaffected so I think the actual breakage is not likely to be so bad.
* I moved the `Lens` utils to a separate module. This is not necessary but it feels cleaner. I’m completely fine with reverting this if you prefer that.
* This is the main part: I added `composeUpdate` which is the analogue of standard function composition but also takes care of combining the effects and I added `zoomUpdate` which lifts an update function into a larger model using a `Lens`. Funnily `zoomUpdate` is actually just `id` + some newtype wrapping and unwrapping :)

In general, it might also be worth considering swapping the type parameters to `Effect`. That would allow us to provide `Functor`, `Applicative` and `Monad` instances. In that case `composeUpdate` would just be `<=<`. I don’t have a strong opinion on whether that’s worth doing.

I’m opening this more as a further starting point of a discussion than something finished so please let me know what you think :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/.stack-work/
+.stack-work/
 result*
 /node_modules/
 dist
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
 .stack-work/
+dist-newstyle/
+cabal.project.local
+.ghc.environment.*
 result*
 /node_modules/
 dist
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ main = startApp App {..}
     subs   = []			  -- empty subscription list
 
 -- | Updates model, optionally introduces side effects
-updateModel :: Action -> Model -> Effect Model Action
+updateModel :: Action -> Model -> Effect Action Model
 updateModel AddOne m = noEff (m + 1)
 updateModel SubtractOne m = noEff (m - 1)
 updateModel NoOp m = noEff m
--- a/examples/compose-update/Main.hs
+++ b/examples/compose-update/Main.hs
@@ -1,0 +1,105 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TupleSections #-}
+module Main where
+
+-- This example demonstrates how you can split your update function
+-- into separate update functions for parts of your model and then
+-- combine them into a single update function operating on the whole
+-- model which combines their effects.
+
+import Control.Monad
+import Data.Monoid
+
+import Miso
+import Miso.Lens
+
+-- In this slightly contrived example, our model consists of two
+-- counters. When one of those counters is incremented, the other is
+-- decremented and the other way around.
+type Model = (Int, Int)
+
+data Action
+  = Increment
+  | Decrement
+  | NoOp
+  deriving (Show, Eq)
+
+-- We are going to use 'Lens'es in this example. Since @miso@ does not
+-- depend on a lens library we are going to define a couple of
+-- utilities ourselves. We recommend that in your own applications,
+-- you depend on a lens library such as @lens@ or @microlens@ to get
+-- these definitions.
+
+-- | You can find this under the same name in @lens@ and
+-- @microlens@. @lens@ also provides the infix operator '%%~' as a
+-- synonym for 'traverseOf'.
+--
+-- In this example we are only going to use this when applied to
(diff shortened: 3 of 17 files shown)


dmjio:
Great work @cocreature :)

> Effect has been generalized to allow an arbitrary number of actions instead of just 0 and 1. This is a breaking change but the smart constructors are unaffected so I think the actual breakage is not likely to be so bad.

Nice, sounds like a transparent change.

> I moved the Lens utils to a separate module. This is not necessary but it feels cleaner. I’m completely fine with reverting this if you prefer that.

+1 for `Miso.Lens`, good refactoring. 

> This is the main part: I added composeUpdate which is the analogue of standard function composition but also takes care of combining the effects and I added zoomUpdate which lifts an update function into a larger model using a Lens. Funnily zoomUpdate is actually just id + some newtype wrapping and unwrapping :)

Do we think a smart constructor for `zoomUpdate` makes sense?

```haskell
(~>) lens model update = over lens update model                                                                                            

newEffect = _1 ~> model $ \x -> x + 1                                                                                                      
  where                                                                                                                                    
    model = (1,1) 
```

I'm kind of reticent to add all these helpers without first letting people build large apps and deeming it necessary. And also, at what point should the onus be on the user to create these niceties.

> In general, it might also be worth considering swapping the type parameters to Effect. That would allow us to provide Functor, Applicative and Monad instances. In that case composeUpdate would just be <=<. I don’t have a strong opinion on whether that’s worth doing.

Flipping the type parameter sounds reasonable (the order of them now does seem backwards). Making `Effect` a `Monad` I've tried before... Correct me if I'm wrong, but I believe that this would allow model updates to be performed in `IO` (this might not longer be true). If so, this kind of violates purity. Also, If we do this, people will want to make it a transformer, which doesn't thrill me.

cocreature:
> Flipping the type parameter sounds reasonable (the order of them now does seem backwards). Making Effect a Monad I've tried before... Correct me if I'm wrong, but I believe that this would allow model updates to be performed in IO (this might no longer be true). If so, this kind of violates purity. Also, If we do this, people will want to make it a transformer, which doesn't thrill me.

Luckily this is not true! The instances would be exactly the same as for `([IO act],)`. Actions will be concatenated using the monoid instance of `[]` but you can’t depend on them using `>>=`. I’m not sure if a transformer is possible (I’ve never seen one for tuple) but I agree that even if it would be, it’s probably not a good idea.

If you are fine with swapping the type parameters then things are a bit different. `composeUpdate` is now redundant since it’s literally just `<=<`. `focusUpdate` is just `id` or if you want a better name or an infix operator, lens libraries provide `traverseOf` and `%%~`. In that case I would thus refrain from adding either of these functions and instead just write an example which demonstrates how you can use these functions to compose update functions.

dmjio:
> Luckily this is not true! The instances would be exactly the same as for ([IO act],). Actions will be concatenated using the monoid instance of [] but you can’t depend on them using >>=. I’m not sure if a transformer is possible (I’ve never seen one for tuple) but I agree that even if it would be, it’s probably not a good idea.

Sounds interesting, what would happen to the model in the bind instance?

> If you are fine with swapping the type parameters then things are a bit different. composeUpdate is now redundant since it’s literally just <=<. focusUpdate is just id or if you want a better name or an infix operator, lens libraries provide traverseOf and %%~. In that case I would thus refrain from adding either of these functions and instead just write an example which demonstrates how you can use these functions to compose update functions.

Yes, this sounds good. I'd like to only provide an interface that is able to be used by lenses, without providing the lenses ourselves, or compose functions. Letting users rely on instead `microlens`.

cocreature:
We’ve already discussed this in private but for the sake of anybody stumbling upon this issue, the `Monad` instance is pretty much identical to the one for [tuples](https://hackage.haskell.org/package/base-4.10.0.0/docs/src/GHC.Base.html#line-344). I’ve now removed `composeUpdate` and `focusUpdate` and instead swapped the type parameters of `Effect` and added `Functor`, `Applicative` and `Monad` instances. I’ll try to get to writing an example demonstrating how to use this tomorrow.

cocreature:
Alright, I’ve just added the example so I think this can be merged.

dmjio:
Fantastic work.

PR #234, opened 2017-08-04
Remove ToMisoString constraint on text

dmjio:
Will get around having to `pack` everything.
- Re-exported `Data.Monoid` from `Miso.String`
- Removed silly `ToMisoString Value` instance.
- Added `ms` function as a concise `toMisoString`

cc @cocreature, can I get a review here.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -1,7 +1,9 @@
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 module Main where
 
 import Miso
+import Miso.String
 
 type Model = Int
 
@@ -31,6 +33,7 @@ data Action
 viewModel :: Int -> View Action
 viewModel x = div_ [] [
    button_ [ onClick AddOne ] [ text "+" ]
- , text (show x)
+ , text $ ms (show x)
  , button_ [ onClick SubtractOne ] [ text "-" ]
  ]
+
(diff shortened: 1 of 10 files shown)


ghcjs-src/Miso/String.hs
```diff
@@ -45,11 +47,12 @@ instance FromJSON MisoString where
 -- | Convenience class for creating `MisoString` from other string-like types
 class ToMisoString str where
   toMisoString :: str -> MisoString
+  ms :: str -> MisoString
```
cocreature: What’s the difference between `toMisoString` and `ms`? Is `ms` just a shorter name? If so, should it really be a class method that instances can potentially override?

exe/Main.hs
```diff
@@ -30,7 +32,8 @@ data Action
 
 viewModel :: Int -> View Action
 viewModel x = div_ [] [
-   button_ [ onClick AddOne ] [ text "+" ]
- , text (show x)
+   button_ [ onClick AddOne ] [ text "foo" ]
```
cocreature: That’s probably not an intentional change.

dmjio: Indeed :)

ghcjs-src/Miso/String.hs
dmjio: Ah, great point. Will factor out into standalone `ms :: ToMisoString s => s -> MisoString`

dmjio:
Should be fixed now.

PR #237, opened 2017-08-06
Geolocation API

dmjio:
cc @cocreature

### DIFF
--- a/ghcjs-src/Miso/Subscription/Geolocation.hs
+++ b/ghcjs-src/Miso/Subscription/Geolocation.hs
@@ -1,0 +1,83 @@
+{-# LANGUAGE RecordWildCards   #-}
+{-# LANGUAGE OverloadedStrings #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Miso.Subscription.Geolocation
+-- Copyright   :  (C) 2016-2017 David M. Johnson
+-- License     :  BSD3-style (see the file LICENSE)
+-- Maintainer  :  David M. Johnson <djohnson.m@gmail.com>
+-- Stability   :  experimental
+-- Portability :  non-portable
+----------------------------------------------------------------------------
+module Miso.Subscription.Geolocation
+  ( geoSub
+  , defGeoOptions
+  , GeoOptions (..)
+  , Geo (..)
+  ) where
+
+import GHCJS.Foreign.Callback
+import GHCJS.Marshal
+import GHCJS.Types
+import JavaScript.Object
+import JavaScript.Object.Internal
+
+import Miso.Html.Internal         ( Sub )
+import Miso.String
+
+data GeoOptions = GeoOptions
+  { enableHighAccuracy :: Bool
+  , maximumAge :: Int
+  , timeout :: Int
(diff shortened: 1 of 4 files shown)


ghcjs-src/Miso/Subscription/Geolocation.hs
```diff
@@ -0,0 +1,83 @@
...
+  = GeoError Code Message
+  | GeoCoords Latitude Longitude
+  deriving (Show, Eq)
+
+type Code = Int
+type Message = MisoString
+type Latitude = Double
+type Longitude = Double
+
+geoSub :: GeoOptions -> (Geo -> action) -> Sub action model
```
cocreature: This should have some docs linking to https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/watchPosition to make it clear when exactly this subscription will fire.

ghcjs-src/Miso/Subscription/Geolocation.hs
```diff
@@ -0,0 +1,83 @@
...
+    pure j
+
+defGeoOptions :: GeoOptions
+defGeoOptions = GeoOptions True 30000 27000
+
+-- | GeoLocation API
+-- https://developer.mozilla.org/en-US/docs/Web/API/Geolocation
+data Geo
+  = GeoError Code Message
+  | GeoCoords Latitude Longitude
```
cocreature: IMHO this should include all the information available in a [Position object](https://developer.mozilla.org/en-US/docs/Web/API/Position) so also the timestamp. Maybe some bang patterns here might also be a good idea. A link to the docs for `Position` would probably also be good.

ghcjs-src/Miso/Subscription/Geolocation.hs
```diff
@@ -0,0 +1,83 @@
...
+    toJSVal enableHighAccuracy >>= \e ->
+      setProp "enableHighAccuracy" e o
+    toJSVal maximumAge >>= \m ->
+      setProp "maximumAge" m o
+    toJSVal timeout >>= \t ->
+      setProp "timeout" t o
+    pure j
+
+defGeoOptions :: GeoOptions
+defGeoOptions = GeoOptions True 30000 27000
```
cocreature: The docs should probably mention what the defaults are. Otherwise the only way to figure this out is to look at the source.

ghcjs-src/Miso/Subscription/Geolocation.hs
```diff
@@ -0,0 +1,83 @@
...
+import GHCJS.Foreign.Callback
+import GHCJS.Marshal
+import GHCJS.Types
+import JavaScript.Object
+import JavaScript.Object.Internal
+
+import Miso.Html.Internal         ( Sub )
+import Miso.String
+
+data GeoOptions = GeoOptions
```
cocreature: The haddocks here should link to the docs for [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions). Personally I would also make the fields strict but I don’t have a particular strong opinion in this case.

dmjio:
Might be better just to make this an example.

PR #238, opened 2017-08-06
Various haddock improvements

cocreature:
Here’s a first attempt at improving the haddocks. I’m by no means done making improvements but I’d rather merge reasonably small improvements than trying to finish it all at once.

### DIFF
--- a/ghcjs-src/Miso/Html/Internal.hs
+++ b/ghcjs-src/Miso/Html/Internal.hs
@@ -72,10 +72,15 @@ import           Miso.Event.Types
 import           Miso.String
 import           Miso.FFI
 
--- | Type def for constructing event subscriptions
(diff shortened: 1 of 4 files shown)


ghcjs-src/Miso/Html/Internal.hs
```diff
@@ -72,10 +72,14 @@ import           Miso.Event.Types
 import           Miso.String
 import           Miso.FFI
 
--- | Type def for constructing event subscriptions
-type Sub a m = IO m -> (a -> IO ()) -> IO ()
+-- | Type synonym for constructing event subscriptions.
+--
+-- The first argument passed to a subscription provides a way to
+-- access the current value of the model. The callback is used to
```
dmjio: Might be worthwhile to add `".. access the current value of the model in a non-blocking fashion"`, or something to ensure the user it's a non-blocking read.

PR #248, opened 2017-08-10
Fix build with servant >= 0.10

cocreature:
These fixes are necessary for servant >= 0.10 and >= 0.11. The
alternative to the CPP would be to just drop support for servant <
0.11 but since until recently servant-0.9 was still the version in
stackage LTS (all snapshots < LTS 9.0), I think it makes sense to
support it a bit longer. nixpkgs unstable has also moved to servant-0.11.

### DIFF
--- a/examples/haskell-miso.org/server/Main.hs
+++ b/examples/haskell-miso.org/server/Main.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE DataKinds                  #-}
 {-# LANGUAGE TypeOperators              #-}
 {-# LANGUAGE TypeApplications           #-}
+{-# LANGUAGE CPP                        #-}
 module Main where
 
 import           Common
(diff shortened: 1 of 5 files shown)


dmjio:
Nice, I'll need to check if our stack files work with the latest resolvers. If so, I'd feel safe dropping support for all pre `0.11` servant, and just instruct users to upgrade. I have a feeling ghcjs 7.10.3 won't though.

cocreature:
> Nice, I'll need to check if our stack files work with the latest resolvers. If so, I'd feel safe dropping support for all pre 0.11 servant, and just instruct users to upgrade. I have a feeling ghcjs 7.10.3 won't though.

IMHO the additional work required on miso’s side to support `servant < 0.11` is sufficiently small that it’s just not worth the potential trouble to drop support now.

dmjio:
Sure, I guess I just would like to avoid CPP in the examples if we could.

dmjio:
We can remove the CPP later if we deprecate usage of `servant < 0.10`.

PR #249, opened 2017-08-10
add IsString instance for views

adnelson:
I'm not sure if there was a reason this instance wasn't defined, but it's pretty trivial to write and at least for me seems to improve the ergonomics.


dmjio:
Brilliant idea. Thank you.

Issue #255, opened 2017-08-12
initialAction doesn't seem to be run (immediately)


FPtje:
## Description
The `startApp` function looks a lot like Elm's. I love that. The `initialAction` field doesn't seem to do anything, though, that is until some other action is run.

## Meta info
Version: `0.7.0.0`

## How to reproduce
Imports and extensions may be a bit off. 

```haskell
{-# LANGUAGE OverloadedStrings #-}

module Main where

import Miso
import Miso.String

main :: IO ()
main = startApp App
    { initialAction = DoThing
    , model  = initialModel
    , update = updateModel
    , view   = viewModel
    , events = defaultEvents
    , subs   = []
    }

initialModel :: MisoString
initialModel = "DoThing was NOT done yet!!!"

data Action
   = DoThing
   | NoOp

updateModel :: Action -> MisoString -> Effect Action MisoString
updateModel NoOp m = noEff m
updateModel DoThing _m = "DoThing was done!!!" <# do
  putStrLn "yeah! Done!"
  pure NoOp

viewModel :: MisoString -> View Action
viewModel m = div_ []
    [ text m
    , button_ [ onClick DoThing ] [ text "Click me!" ]
    ]

```

## Expected behaviour
Page shows `DoThing was done!!!` right after loading, and prints `yeah! Done!` in console.

## Actual behaviour
Page shows `DoThing was NOT done yet!!!` indefinitely until the button is pressed. The `yeah! Done!` is also not printed.

Oddly, when you *do* press the button, the `putStrlLn` will print twice. Pressing the button a second time will have it print once. 

I think this is just an issue of  the `initialAction` being delayed until the next action is run, rather than the `initialAction` being forgotten entirely.

dmjio:
Fixed in #256

PR #256, opened 2017-08-12
Async effects

dmjio:
This PR performs twofold:
  - Ensures we notify the redrawer only when we are certain an action exists. Fixing #255  
  - Evaluates effects asynchronously. Previously, evaluating accumulated `effects` was blocking redrawing.

cc @cocreature @FPtje


cocreature:
lgtm.

PR #257, opened 2017-08-13
Add link tag

ptigwe:
Comes in handy, especially when previewing with ghci.


dmjio:
@ptigwe, Awesome! Thanks!

PR #261, opened 2017-08-14
Support for generic direction configurations

ptigwe:
I added some support for generic direction configurations when subscribing to the keyboard. ``arrowsSub`` is still available and works as expected. I've also added ``wasdSub`` to provide a similar functionality but with the 'wasd' keys.

Subscribing to both ``arrowsSub`` and ``wasdSub`` is not stable, hence the third method of subscription ``directionSub`` to achieve this.

``directionSub ([38, 87], [40, 83], [37, 65], [39, 68])``

### DIFF
--- a/ghcjs-src/Miso/Subscription/Keyboard.hs
+++ b/ghcjs-src/Miso/Subscription/Keyboard.hs
@@ -16,7 +16,9 @@ module Miso.Subscription.Keyboard
     Arrows (..)
     -- * Subscriptions
   , arrowsSub
+  , directionSub
   , keyboardSub
+  , wasdSub
   ) where
 
 import           Data.IORef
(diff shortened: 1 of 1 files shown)


dmjio:
@ptigwe great idea. All for abstracting out common functionality to provide additional arrow subscriptions.

One suggestion, as opposed to using `[Int]`, could we simply use `Int`? Just trying to remove one level of indirection if possible. Otherwise, let me know if there's something I'm missing regarding the usage of a list.

So from
```haskell
toArrows :: ([Int], [Int], [Int], [Int]) -> Set Int -> Arrows
toArrows (up, down, left, right) set =
  Arrows {
    arrowX =
      case (check left, check right) of
        (True, False) -> -1
        (False, True) -> 1
        (_,_) -> 0
  , arrowY =
      case (check down, check up) of
        (True, False) -> -1
        (False, True) -> 1
        (_,_) -> 0
  }
  where
    check = any (`S.member` set)
```

to this
```haskell
toArrows :: (Int, Int, Int, Int) -> Set Int -> Arrows
toArrows (up, down, left, right) set =
  Arrows {
    arrowX =
      case (check left, check right) of
        (True, False) -> -1
        (False, True) -> 1
        (_,_) -> 0
  , arrowY =
      case (check down, check up) of
        (True, False) -> -1
        (False, True) -> 1
        (_,_) -> 0
  }
  where
    check = (`S.member` set)
```

ptigwe:
If one tries to subscribe separately to both the *arrow keys* and *wasd* using the same action, then you will run into some issues with the directions getting set and unset in rapid successions (or probably not even getting set at all). 

One way around this is to use ``[Int]`` to specify the key bindings (e.g. ``([38, 87], [40, 83], [37, 65], [39, 68])``) to have only a single subscription which works properly with a single type of action. 

While another method would be to use two different actions and ensure they get merged properly within the app.

dmjio:
Alright, I'm convinced.  This patch has been merged manually. I rebased some things on top of your branch, specifically defining `wasdSub` and `keyboardSub` in terms of `directionSub`
```haskell
wasdSub :: (Arrows -> action) -> Sub action model
wasdSub = directionSub ([87], [83], [65], [68])
```
This merge has retained your attribution and commit history, located [here](https://github.com/haskell-miso/miso/commit/361687c8ab6c6c2bc727937e105448c8fe2ee37e). Thanks again for the great work and your contributions. I'm excited for `2048` :)

PR #263, opened 2017-08-17
Add 2048

dmjio:
cc @ptigwe


dmjio:
I don't mind merging now, and then when you get those sexy css transitions in, we'll bump the rev on the nix build.

PR #271, opened 2017-08-22
Propagate to parent event handlers when no event present

dmjio:
cc @FPtje, have tested this on your example with success, along with examples.


FPtje:
Nice! Thanks again for the quick fix!

dmjio:
For sure 🍻

Issue #274, opened 2017-08-24
Decoding list


ptigwe:
We need to be able to decode lists of objects so as to be able to handle nested information about events (e.g. [Touch events](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events)).
Related to #273 and #121.

ptigwe:
Resolved #275.

Issue #273, opened 2017-08-24
Decoding multiple fields


ptigwe:
Certain events, especially [Touch events](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events) have a bit of a nested nature. At the moment, with the use of `decodeAt :: [JSString]` we are only capable of looking at a single one of such nested objects. One possibility involves changing the type of `decodeAt` to `[[JSString]]`.
Related to #121.

ptigwe:
Resolved #275.

dmjio:
Thanks again @ptigwe, great work

Issue #121, opened 2017-06-02
Touch events


dmjio:
Will probably need to implement them as a subscription. `touchstart`, `touchend`, etc.

ptigwe:
've looked into this a bit. As one might expect, creating the listeners is quite trivial. The main issue comes when you try to decode the event. Unlike other events (like mouse and keyboard), with touch events, the data resides within objects nested within the event object. I attempted modifying the `objectToJSON` function so as to extract some of this info, which is where I'm currently stuck at. I've tried but not so sure as to how to modify the `jsvalToValue` function so that it is capable of parsing nested objects. I keep getting `TypeError: Cannot read property 't' of undefined`. Not really sure how to proceed.

```javascript
function objectToJSON (at, obj) {
  for (var i in at) obj = obj[at[i]];
  console.log(obj);
  var newObj = objToJSON(obj, 2);
  console.log(newObj);
  return (newObj);
}

function objToJSON (obj, rec) {
  var newObj = {}
  for (var i in obj) {
    if (typeof obj[i] == "string" || typeof obj[i] == "number" || typeof obj[i] == "boolean")
      newObj[i] = obj[i];
    else if (typeof obj[i] == "object" && rec != 0 && i == "changedTouches") {
      newObj[i] = objToJSON(obj[i], rec - 1);
    }
  }
  return (newObj);
}```

**Comment by ptigwe** [comment]
Thanks to @dmjio, the following code resolves this issue and enables reading the first touch event. In the situation where only a single touch is required, the following code should suffice.

```haskell
touchDecoder :: Decoder Touch
touchDecoder = Decoder {..}
  where
    decodeAt = ["touches", "0"]
    decoder = withObject "touch" $ \o -> 
          Touch <$> o .: "clientX" <*> o .: "clientY" 

onTouchMove :: (Touch -> action) -> Attribute action
onTouchMove = on "touchmove" touchDecoder
```

dmjio:
awesome, 👌

PR #275, opened 2017-08-28
List of Objects and Multi-target Decoding

ptigwe:
As this is a proof of concept, the implementation is quite limited (at the time of initial commit, only the SVG example works, others are most likely broken). As the Haskell code hasn't changed much, parts of the functionality are being simulated by the `objectsToJSON` function.

The general plan is to change the type of `decodeAt` to `[[MisoString]]`. This will allow for the specification of multiple targets from which to retrieve information. This gets passed to `objectsToJSON`, which returns an array of objects in the same order specified within `decodeAt` by making multiple calls to `objectToJSON`.

As a slightly related, can anyone see if they are still getting the error which @FPtje reported in ghcjs/ghcjs-base#88. At the moment, I'm dealing with this using a temporary fix I got from ghcjs/shims#41, deleting the second `h$listProps` function upon every build.

Related: #121 #274 #273

### DIFF
--- a/examples/svg/Main.hs
+++ b/examples/svg/Main.hs
@@ -9,6 +9,7 @@ import qualified Data.Map           as M
 import           Miso
 import           Miso.String        (MisoString, pack, ms)
 import           Miso.Svg           hiding (height_, id_, style_, width_)
+import Touch
 
 main :: IO ()
 main = startApp App {..}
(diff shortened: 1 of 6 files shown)


dmjio:
@ptigwe Also, might be nice to make a special combinator for constructing `[[MisoString]]`.

ptigwe:
@dmjio The current version doesn't require much updates to existing code which make use of the smart constructors for Decoders. The main change in the Haskell code is the definition of `DecoderTarget`.

PR #276, opened 2017-08-28
Added more info to URI returned by getCurrentURI

FPtje:
# Motivation
I needed the URL fragment in the app I'm building. Found out they were missing 🍜

# Technical details
This PR fills in `uriScheme` and `uriFragment`

They were set to empty strings, but they can easily be retrieved from `window.location`, like `uriQuery` and `uriPath` are.

Two notes:

- I didn't touch `uriAuthority`, because I'm not sure how to construct it.
- The value filled in for `uriPath` strips the first `/` character, which seems to oppose the Haddock of [URI](http://hoogle.lumi.guide.local/file/nix/store/qw0lvqkyl0gicmhdwmfc26ari8dyqdg3-network-uri-2.6.1.0/share/doc/x86_64-linux-ghc-8.0.2/network-uri-2.6.1.0/html/Network-URI.html#t:URI). I didn't touch this because I have a gut feeling that there's a really good reason for this.

### DIFF
--- a/ghcjs-src/Miso/Subscription/History.hs
+++ b/ghcjs-src/Miso/Subscription/History.hs
@@ -41,11 +41,11 @@ getCurrentURI = getURI
 getURI :: IO URI
 {-# INLINE getURI #-}
 getURI = do
-  URI <$> pure mempty
+  URI <$> do unpack <$> getProtocol
       <*> pure Nothing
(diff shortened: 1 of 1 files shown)


dmjio:
Fantastic! Thanks!

PR #282, opened 2017-09-05
Add BangPatterns to foldEffects

cocreature:
Tuples are lazy so without this change the model will never actually
be forced in the foldl' (foldEffects …) which could lead to spaceleaks.

### DIFF
--- a/ghcjs-src/Miso.hs
+++ b/ghcjs-src/Miso.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE RecordWildCards     #-}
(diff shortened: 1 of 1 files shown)


dmjio:
Can you confirm this indeed does cause a space leak. Or create a scenario where this behavior is exhibited. I'd rather be lazy (no pun intended) and not add additional computation if we can't confirm it's a leak.

cocreature:
Take a model that just consists of an integer and a bunch of actions that increment that integer. Then you’ll have a spaceleak in https://github.com/haskell-miso/miso/blob/43ffd43830f24f6506f1a5d2e358c9d0ea5eb3f0/ghcjs-src/Miso.hs#L82 despite the `foldl'` which is probably meant to prevent this exact thing from happening. If you really want to be lazy in this case (I don’t see what you gain by that, if your model has a field that you want to compute lazily you’ll just not make that field strict) then that should be at least a `foldr`. A `foldl'` that is not sufficiently strict is not useful.

PR #285, opened 2017-09-13
README.md: Update Sample Application to reflect latest sample-app/Main.hs

patrickmn:
The current one doesn't compile because Miso.String isn't imported.


dmjio:
Thanks! 🍻

PR #289, opened 2017-09-17
Fixed websockets not reopening properly after being closed

### LINKED ISSUE #288: Websockets: non-working state after socket closes
# How to reproduce

- Take the [websocket](https://github.com/haskell-miso/miso#websocket) example.
- Change the URL to some websocket server that sends one message and then immediately closes the socket

# Expected behaviour
- That one message is printed
- The web socket is closed
- [This here logic](https://github.com/haskell-miso/miso/blob/master/ghcjs-src/Miso/Subscription/WebSocket.hs#L74-L80) re-opens the socket within three seconds
- repeat step 1

One could argue that a neatly closed socket shouldn't be re-opened like that, but that's not the point of this issue.

# Actual behaviour
- That one message is printed
- The web socket is closed
- The socket is re-opened within three seconds
- No messages about this socket arrive to the application

# Why it happens
The `onOpen`, `onMessage`, `onClose` and `onError` handlers [here](https://github.com/haskell-miso/miso/blob/master/ghcjs-src/Miso/Subscription/WebSocket.hs#L81-L97) refer to the `socket` defined on [line 71](https://github.com/haskell-miso/miso/blob/master/ghcjs-src/Miso/Subscription/WebSocket.hs#L71). That's the *old* socket, not the *new* `socket` created on [line 80](https://github.com/haskell-miso/miso/blob/master/ghcjs-src/Miso/Subscription/WebSocket.hs#L80) when the connection closes.

# Solution
Not sure, but go in recursion on line 79-80 instead of rewriting the `IORef`?

I haven't made a PR for this because I'm not sure whether recreating the socket on close is *always* a good idea. I reckon it's a good idea to reconnect when the socket closed due to an error, but not when it's closed properly.

# Simple example server
Not full code, sorry for that

```haskell
import qualified "wai" Network.Wai as Wai
import qualified "wai-websockets" Network.Wai.Handler.WebSockets as WebSockets
import qualified "websockets" Network.WebSockets.Connection as WebSockets

main = WebSockets.websocketsOr opts serverApp nonSocket
  where
    serverApp :: WebSockets.PendingConnection -> IO ()
    serverApp pconn = do
            conn <- WebSockets.acceptRequest pconn
            WebSockets.sendTextData conn ("\"AAA\"" :: T.Text)

    nonSocket _req respond =
            respond $ Wai.responseLBS
                Http.notAcceptable406
                [("Content-Type", "text/plain")]
                "This is a websocket route. Connect to it using websockets."

    opts = WebSockets.ConnectionOptions $ pure ()
```

FPtje:
Fixes #288 

I haven't changed the behaviour of sockets that closed cleanly. This PR just fixes the re-opening of sockets after closing. See #288 for more detail.

Tested locally with my own application.

Edit: ~~Wait, hold on, I didn't think the threading through. Now it's forking every three seconds~~
Fixed

### DIFF
--- a/ghcjs-src/Miso/Subscription/WebSocket.hs
+++ b/ghcjs-src/Miso/Subscription/WebSocket.hs
@@ -67,17 +67,10 @@ websocketSub
   -> Protocols
   -> (WebSocket m -> action)
   -> Sub action model
-websocketSub (URL u) (Protocols ps) f _ sink = do
+websocketSub (URL u) (Protocols ps) f getModel sink = do
   socket <- createWebSocket u ps
   writeIORef websocket (Just socket)
-  --  Handle Reconnects
-  void . forkIO . forever $ do
-    threadDelay (secs 3)
-    Just s <- readIORef websocket
-    status <- getSocketState' s
-    when (status == 3) $ do
-      atomicWriteIORef websocket
-        =<< Just <$> createWebSocket u ps
+  void . forkIO $ handleReconnect
   onOpen socket =<< do
     asyncCallback $ do
       sink (f WebSocketOpen)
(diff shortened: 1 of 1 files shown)


dmjio:
fixed in https://github.com/haskell-miso/miso/pull/291

PR #291, opened 2017-09-21
Fixes websocket reconnection logic

dmjio:
@FPtje can I get a review here.

Rebased your code with one additional commit to reconnect only when the socket has closed abnormally (when `CLOSED_NORMAL` is not the case). 

Note, this doesn't include client-side closing or reconnecting (just server). Handling client side socket closing / reopening goes back to managing dynamic subscriptions, which is something `miso` doesn't have at current.

### DIFF
--- a/ghcjs-src/Miso/Subscription/WebSocket.hs
+++ b/ghcjs-src/Miso/Subscription/WebSocket.hs
@@ -57,6 +57,10 @@ websocket :: IORef (Maybe Socket)
 {-# NOINLINE websocket #-}
 websocket = unsafePerformIO (newIORef Nothing)
 
+closedCode :: IORef (Maybe CloseCode)
+{-# NOINLINE closedCode #-}
+closedCode = unsafePerformIO (newIORef Nothing)
+
 secs :: Int -> Int
 secs = (*1000000)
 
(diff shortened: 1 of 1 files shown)


FPtje:
Looks good. I reckon the normal closing of a websocket is indeed the only time it shouldn't reconnect.

If you like I can run a quick test in a couple of hours.

dmjio:
@FPtje that would be fantastic.

FPtje:
Looks like it's still recreating the sockets. Hold on, let me see if I can do some debugging.

FPtje:
The server isn't sending a nice close message. I'll have to postpone debugging

FPtje:
Finally I got the server to work right. It all works properly. The server sends the close message, after which Miso closes the web socket. Miso doesn't recreate the socket.

dmjio:
@FPtje I'd be curious to know why it wasn't working initially. `sendClose` from websockets /should/ send the correct code by default.
```haskell
sendClose :: WebSocketsData a => Connection -> a -> IO ()
sendClose conn = sendCloseCode conn 1000
```
We can merge for now and add more sophisticated reconnection logic later. Thanks again!

FPtje:
It wasn't working initially because I didn't call `sendClose` :)

dmjio:
:)

PR #293, opened 2017-09-24
Initial commit of syncChildren

dmjio:


jsbits/diff.js
```diff
@@ -165,30 +168,34 @@ function syncChildren (os, ns, parent) {
...
 	   -> [ a b c ] <- new children
 	*/
 	if (oldFirstIndex > oldLastIndex) {
-	    var node = createNodeDontAppend(nFirst);
+	    createElement(nFirst);
 	    /* insertBefore's semantics will append a node if the second argument provided is `null` or `undefined`.
 	       Otherwise, it will insert node.domRef before oLast.domRef. */
-	    parent.insertBefore(node.domRef, oLast ? oLast.domRef : null);
-	    os.splice(newFirstIndex, 0, nFirst);
+            parent.insertBefore (nFirst.domRef, os[oldLastIndex] ? os[oldLastIndex].domRef.nextSibling : null);
```
cocreature: For consistency, this should either use `oLast`  or `oLast` should be removed.

jsbits/diff.js
```diff
@@ -198,168 +205,152 @@ function syncChildren (os, ns, parent) {
 	    tmp = oldLastIndex - oldFirstIndex;
 	    while (tmp >= 0) {
 	      parent.removeChild(os[oldFirstIndex].domRef);
```
cocreature: I might be missing something here but I think this loop should be incrementing `oldFirstIndex`. Otherwise you keep removing the same element.

jsbits/diff.js
```diff
@@ -198,168 +205,152 @@ function syncChildren (os, ns, parent) {
...
-	    newFirstIndex++;
-	    oldFirstIndex++;
-	    continue;
-	} else if (oLast.key === nLast.key) {
-	    newLastIndex--;
-	    oldLastIndex--;
-	    continue;
-	}
+	else if (oFirst.key === nFirst.key || oLast.key === nLast.key) {
+            while (os[oldFirstIndex].key === ns[newFirstIndex].key && os[oldLastIndex].key === ns[newLastIndex].key) {
```
cocreature: Does the order in which elements are diffed actually matter? Otherwise it seems like this loop can be removed since the two loops below cover this case.

jsbits/diff.js
```diff
@@ -198,168 +205,152 @@ function syncChildren (os, ns, parent) {
...
+	   -> [ a b d ] <- new children
+	   becomes
+	   -> [ a b d ] <- old children
+	   -> [ a b d ] <- new children
+	   and now we happy path
+        */
+        else if (oFirst.key === nLast.key) {
+	    parent.insertBefore(oFirst.domRef, oLast.domRef.nextSibling);
+	    /* swap positions in old vdom */
+   	    os.splice(oldLastIndex, 0, os.splice(oldFirstIndex, 1)[0]);
```
cocreature: IMHO this should be wrapped up in some kind of `swap` function. I also suspect that just assigning things to a temporary variable and doing the swap that way is going to be faster (and for me personally it’s easier to read) but I haven’t measured this.

jsbits/diff.js
```diff
@@ -198,168 +205,152 @@ function syncChildren (os, ns, parent) {
...
+           We move d to end of list, mutate old vdom to reflect the change
+	   -> [ b a d ] <- old children
+	   -> [ d b a ] <- new children
+	   becomes
+	   -> [ a b d ] <- old children
+	   -> [ a b d ] <- new children
+	   and now we happy path
+	*/
+        else if (oLast.key === nFirst.key) {
+	    os.splice(oldFirstIndex, 0, os.splice(oldLastIndex, 1)[0]);
```
cocreature: Same here.

jsbits/diff.js
```diff
@@ -198,168 +205,152 @@ function syncChildren (os, ns, parent) {
...
+	       This results in new list below w/ updated index position
+	       -> [ b a e c ] <- old children
+	       -> [ b e a j ] <- new children
+                      ^
+   	    */
+	    if (found) {
+		/* Swap DOM references */
+		parent.insertBefore(node.domRef, oFirst.domRef);
+                /* Move item to correct position */
+        	os.splice(oldFirstIndex, 0, os.splice(tmp, 1)[0]);
```
cocreature: And here.

jsbits/diff.js
dmjio: fancy :)

jsbits/diff.js
dmjio: Yea we probably can remove this case. Doing two at once might be overrated.

jsbits/diff.js
dmjio: agreed

jsbits/diff.js
dmjio:
```javascript
Array.prototype.move = function(from, to) {
    this.splice(to, 0, this.splice(from, 1)[0]);
};
```

jsbits/diff.js
dmjio: Nice catch, I removed mutating `os` since this is a base case, but did not realize that `oldFirstIndex` wasn't getting incremented. Thanks. Will fix.

jsbits/diff.js
cocreature: I still suspect that this is potentially slower than
```js
Array.prototype.move = function(from, to) {
  var tmp = this[to];
  this[to] = this[from];
  this[from] = tmp;
}
```

jsbits/diff.js
```diff
@@ -127,13 +127,16 @@ function diffCss (cCss, nCss, node) {
 }
 
 function hasKey (cs) {
-    return cs && cs[0] && cs[0].key;
+  return cs && cs[0] && cs[0].key;
 }
 
 function diffChildren (cs, ns, parent) {
     var longest = ns.length > cs.length ? ns.length : cs.length;
```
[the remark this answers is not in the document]
dmjio: Nice. Will change.

jsbits/diff.js
dmjio: https://jsperf.com/ it :)

cocreature: It’s not even close, using a temporary variable is clearly faster (the difference is similar in Chrome)
screenshot-2017-9-25 swap two elements in an array jsperf

dmjio: Very nice. I'm convinced. Can you make a PR?

exe/Main.hs
```diff
@@ -2,38 +2,58 @@
...
-updateModel SubtractOne m = noEff (m - 1)
 updateModel NoOp m = noEff m
-updateModel SayHelloWorld m = m <# do
-  putStrLn "Hello World!" >> pure NoOp
+updateModel (SetNums xs) _ = noEff xs
+updateModel (Say x) m = m <# do
+  print x >> pure NoOp
+updateModel NewNums m = m <# do
+  nums <- replicateM 40 getRand
+  print nums >> pure (SetNums nums)
```
cocreature: Fwiw, I don’t think this should replace the counter example. The counter example seems easier to understand and thereby a better choice for the initial example people see. Obviously adding a separate example for this is totally fine.

dmjio: Agreed, I'm going to undo this change.

dmjio:
Replaced by #370

PR #294, opened 2017-09-24
Introduce the Transition monad

basvandijk:
This monad is useful for succinctly expressing model transitions. We use it a lot at LumiGuide. It's especially useful when used in combination with the stateful [lens operators](http://hackage.haskell.org/package/lens-4.15.4/docs/Control-Lens-Operators.html) (all operators ending in "`=`").

@FPtje I removed our `TransitionM` type synonym because I think it adds to much noise.

### DIFF
--- a/ghcjs-src/Miso/Types.hs
+++ b/ghcjs-src/Miso/Types.hs
@@ -9,8 +9,17 @@
 ----------------------------------------------------------------------------
 module Miso.Types
   ( App (..)
+
+    -- * The Transition Monad
+  , Transition
+  , fromTransition
+  , toTransition
+  , scheduleIO
   ) where
 
+import           Control.Monad.Trans.Class (lift)
+import           Control.Monad.Trans.State.Strict (StateT(StateT), execStateT)
+import           Control.Monad.Trans.Writer.Strict (WriterT(WriterT), Writer, runWriter, tell)
 import qualified Data.Map           as M
 import           Miso.Effect
 import           Miso.Html.Internal
@@ -21,7 +30,8 @@ data App model action = App
   { model :: model
   -- ^ initial model
   , update :: action -> model -> Effect action model
-  -- ^ Function to update model, optionally provide effects
+  -- ^ Function to update model, optionally provide effects.
+  --   See the 'Transition' monad for succinctly expressing model transitions.
   , view :: model -> View action
   -- ^ Function to draw `View`
   , subs :: [ Sub action model ]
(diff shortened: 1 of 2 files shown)


FPtje:
This is what I've discussed with @dmjio. Usage example here: https://github.com/FPtje/miso-transition-example

cocreature:
Can you expand a bit on the motivations for this? You mention being able to use the stateful lens operators but you can get pretty much the same using the “normal” lens operators and monadic composition, e.g. for the example you added to the docs (great docs btw!), you could write

```haskell
update MyAction1 = field1 .~ value 1 >=>
                   counter +~ value
update MyAction2 = field2 %~ f >=>
                   scheduleIO $ do
                     putStrLn "Hello"
                     putStrLn "World!"
```
where `scheduleIO = (#>)`.

So is the motivation just do notation over `>=>`? (Not saying that’s a bad motivation, I’m just curious).

That said, the actual change is totally fine with me since it forces nobody to use `Transition`.

dmjio:
@basvandijk, @FPtje Awesome work. This will make a great addition to the external API. Very nice documentation too. `scheduleIO` seems to be a good descriptive name as well, infers to the user that their actions will be evaluated asynchronously. I imagine this API is how most advanced users will work, along with pure (sub)components. Thanks for spearheading the user experience side of things. 

I don't have any recommendations for major changes really, I think it is all good. Only nit I'd have is would it be possible to actually keep `TransitionM` but with swapped type parameters (equivalent to the order that `Effect action model` has). And also to keep `Transition` defined in terms of `TransitionM`, and with hardcoded `()`. Like below.

```haskell
type Transition action model = TransitionM model action ()
type TransitionM action model = StateT model (Writer [IO action])
```

Since this is an "advanced" feature I can see a case for users wanted to transform `ReaderT` over `TransitionM` to hold `action` or some kind of context that has IO actions in it. e.g.

```haskell
type MyThing action model a = ReaderT MyEnv (TransitionM action m) a
```

It seems `mtl` is a new dep. being added. This seems fine since `mtl` is lightweight and is mostly just instances over `transformers` and has a very stable API (`StateT` / `WriterT` won't change).

Once the `miso` O'Reilly book is released we'll have to include `Transition` in the advanced section, and maybe describe a bit about transformers and lenses 😄

basvandijk:
I realise I only depended on `mtl` because of the overloaded [tell](http://hackage.haskell.org/package/mtl-2.2.1/docs/Control-Monad-Writer-Strict.html#v:tell) in `scheduleIO`. I replaced that with "lift . [tell](http://hackage.haskell.org/package/transformers-0.5.4.0/docs/Control-Monad-Trans-Writer-Strict.html#v:tell)" where `tell` is from `transformers` so I dropped the `mtl` dependency.

I've swapped the `model` and `action` parameters of `Transition`. 

Regarding `TransitionM`, note that the current `Transition` was the previous `TransitionM`. So your examples works just fine with `Transition`:

```haskell
type MyThing action model a = ReaderT MyEnv (Transition action model) a
```

Do you have use cases in mind for a `Transition action model ()` without the `()`?

Cool that you're writing a book!

BTW I'm giving a workshop on Nix at the [Haskell eXchange 2017](https://skillsmatter.com/conferences/8522-haskell-exchange-2017#program) next month in London together with [Peter Simons](https://github.com/peti). I'm going to show how to build a web application in Haskell. The focus is not on Haskell but on how to use the Nix infrastructure. However, I'm going to adapt the `miso` todo-app example by giving it a `servant` backend to store todo entries in a database using `opaleye`. Hopefully this will convince a few more people to try out `miso`.

basvandijk:
> Can you expand a bit on the motivations for this?

@cocreature the motivation is mainly that composition is a bit nicer in the `Transition` monad than in the `model -> Effect action model` function. What is also nice is that you can use monadic functions like [when](http://hackage.haskell.org/package/base-4.10.0.0/docs/Control-Monad.html#v:when), [unless](http://hackage.haskell.org/package/base-4.10.0.0/docs/Control-Monad.html#v:unless) or [for_](http://hackage.haskell.org/package/base-4.10.0.0/docs/Data-Foldable.html#v:for_) to optionally update your model or schedule IO actions. IMHO the use of these functions improves readability.

dmjio:
> I realise I only depended on mtl because of the overloaded tell in scheduleIO. I replaced that with "lift . tell" where tell is from transformers so I dropped the mtl dependency.

Fantastic!

> I've swapped the model and action parameters of Transition.

Thanks :) 

> Do you have use cases in mind for a Transition action model () without the ()?

I don't really think so, my thinking was that it might become onerous for the users to continually specify `()` at the end of `Transition` if they don't plan on transforming it. `type Transition a m = StateT m (Writer [IO a]) ()` might make the API a bit cleaner, but specifying `()` would inhibit users from stacking monads.  So for users who wanted to add on a `ReaderT` let's say, they could do so as `type TransitionWithEnv = ReaderT MyEnv TransitionM`. Where `TransitionM` is defined w/o specifying the `()`. 

The current definition of `Transition` is fine as is (without `()`). I suppose we could add `type Trans a m = Transition a m ()` later if we wanted to achieve the same effect as described above. Dangling `()` is more of a nitpick, just wanted to keep the type signatures as close to `Effect` as possible. 

> BTW I'm giving a workshop on Nix at the Haskell eXchange 2017 next month in London together with [Peter Simons](https://github.com/peti)

That sounds like a lot of fun. Give [Peter Simons](https://github.com/peti) my regards. Wish I could make it out there to see it live!

> I'm going to adapt the miso todo-app example by giving it a servant backend to store todo entries in a database using opaleye. Hopefully this will convince a few more people to try out miso.

I bet it will :) Thanks for doing this. Going to go ahead and invite you to the github org as well.
 
Cheers 🍻

PR #298, opened 2017-10-01
Delegation perf

dmjio:
Avoid allocations by reusing arrays.


dmjio:
For the record, this hasn't shown to increase performance. But that might be due to other bottlenecks drowning out the benefits of this PR.

RobertFischer:
I'm not surprised by the result: you're trading fairly short-lived allocations of arrays (allocated locally, GC'd fairly quickly, probably never promoted out of cache or into the older generation) for long-lived allocations plus long-lived overhead (memory contentious, copied about, etc., etc.).

RobertFischer:
What's the test you're running for performance improvements?  I'd like to take a swing at it.

PR #303, opened 2017-10-11
remove redundant calls to notify

adnelson:
`pushStateNoModel` and `replaceTo'` respectively each call `notify chan` as their last step, so there's no need to call `notify chan` a second time when they are called in `pushURI` and `replaceURI`.


dmjio:
nice catch, thanks! 👌

PR #305, opened 2017-10-12
Fix haskell-miso.org navigation menu

brandonhamilton:
Fix highlighting of the navigation menu page names when the URI is changed


dmjio:
❤️

dmjio:
@brandonhamilton Looks great! Thanks again. It's live too.

PR #307, opened 2017-10-21
Remove FailFatal to match servant-server’s behavior

cocreature:
Previously miso made a distinction between fatal errors caused by
failures to parse parameters and non-fatal errors. In the case of
fatal errors miso did not try other routes. However, this does not
match the behavior of servant-server which will try other routes on
parse failures. Since the same routes are used on the server and the
client this can cause problems (and it is also simply confusing for
people coming from servant-server).

Here’s an example to demonstrate `servant-server`’s behavior:
```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TypeOperators #-}
import           Servant
import           Data.Text (Text)
import qualified Data.Text as Text
import           Data.Monoid
import           Network.Wai.Handler.Warp
import           Data.Proxy

type API = "user" :> Capture "user" Int :> Get '[JSON] Text
         :<|> "user" :> "all" :> Get '[JSON] [Text]

server :: Server API
server = userHandler :<|> newUserHandler

userHandler :: Int -> Handler Text
userHandler userId = pure ("user" <> (Text.pack . show) userId)

newUserHandler :: Handler [Text]
newUserHandler = pure ["user42", "user23"]

main :: IO ()
main = run 8080 (serve (Proxy :: Proxy API) server)
```
`servant-server` will choose the second route for `/user/all` while miso will (without the changes in this PR) return a route failure.

### DIFF
--- a/src/Miso/Router.hs
+++ b/src/Miso/Router.hs
@@ -58,12 +58,7 @@ data Location = Location
   } deriving (Show, Eq, Ord)
 
 -- | When routing, the router may fail to match a location.
--- Either this is an unrecoverable failure,
--- such as failing to parse a query parameter,
--- or it is recoverable by trying another path.
-data RoutingError
-  = Fail
-  | FailFatal
+data RoutingError = Fail
   deriving (Show, Eq, Ord)
 
 -- | A 'Router' contains the information necessary to execute a handler.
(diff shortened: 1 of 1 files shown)


dmjio:
Thanks.

PR #309, opened 2017-10-25
Handle consecutive text nodes when parsing server-rendered DOM

cocreature:


jsbits/isomorphic.js
```diff
@@ -1,20 +1,47 @@
...
+  vtree.domRef = node;
+  for (i = 0; i < vtree.children.length; i++) {
+    vdomChild = vtree.children[i];
+    domChild = node.childNodes[j];
+    if (vdomChild.type === "vtext") {
+      if (!inTextBlock) {
+        // This is the first text node in a block of consecutive text nodes.
+        inTextBlock = true;
+        if (domChild.nodeType !== Node.TEXT_NODE) {
+          console.error("miso error: Expected text node but found got element");
```
adnelson: just printing a message here doesn't seem beneficial; if it's an error shouldn't we stop?

jsbits/isomorphic.js
```diff
@@ -1,20 +1,47 @@
...
+      } else {
+        // In this case there have been other text nodes directly before this node.
+        // There is thus no corresponding DOM node and we need to create a new node.
+        vdomChild.domRef = document.createTextNode("");
+        vdomChild.text = "";
+      }
+    } else {
+      inTextBlock = false;
+      if (domChild.nodeType !== Node.ELEMENT_NODE) {
+        console.error("miso error: Expected element but got text node");
```
adnelson: similarly here

jsbits/isomorphic.js
```diff
@@ -1,20 +1,47 @@
...
+  var i = 0,
+    j = 0,
+    // This indicates if we are in a block of consecutive text nodes.
+    inTextBlock = false,
+    vdomChild,
+    domChild;
+  vtree.domRef = node;
+  for (i = 0; i < vtree.children.length; i++) {
+    vdomChild = vtree.children[i];
+    domChild = node.childNodes[j];
```
adnelson: I'm getting an error where `domChild` evaluates to `undefined` because the node has no children. Although I'm not sure my error is related to what's going on in this function (yet), I think we need to account for the possibility that `domChild` is undefined...

cocreature: That error message is what should be fixed by this PR so I don’t think we need to account for this.

jsbits/isomorphic.js
cocreature: Good point, fixed!

jsbits/isomorphic.js
adnelson: Hmm, so after a little more investigation the error I was seeing was due to the fact that my `View` started with a `body_` rather than with some internal element (like `div_` or whatever). Because of this, the `vtree` object in `copyDOMIntoVTree` was meant to represent the `body` tag instead of the first child of the `body` tag. So basically replacing `document.body.firstChild` with `document.body` fixed the issue. I'm not sure if that's in the scope of this PR, but I wanted to get it out there. Maybe we could do something like

```javascript
function copyDOMIntoVTree(vtree) {
  var elem = vtree.tag === "body" ? document.body : document.body.firstChild;
  walk(vtree, document.body); 
}
```

Unfortunately I don't have enough knowledge of exactly what the `walk` function is *supposed* to be doing, to know if that's the right move.

cocreature: It’s definitely out of scope for this PR but it’s also the wrong solution. You can only render inside of `body` with miso so this is a user error. We might consider outputting a warning in that case but trying to fix it is the wrong solution imho. Either way, let’s move discussions about things unrelated to this PR to separate issues.

dmjio: As of now we can render on any DOM element.

jsbits/isomorphic.js
```diff
@@ -1,20 +1,52 @@
...
+    if (vdomChild.type === "vtext") {
+      if (!inTextBlock) {
+        // This is the first text node in a block of consecutive text nodes.
+        inTextBlock = true;
+        if (domChild.nodeType !== Node.TEXT_NODE) {
+          throw new MisoException("Expected text node but found got element");
```
FPtje: The word "got" seems to be a leftover from a rewrite of the error message.

FPtje:
This PR can be closed, as a later version of this has been merged.

PR #311, opened 2017-11-08
Make topnav work with JS disabled. Closes #241

### LINKED ISSUE #241: Browsing haskell-miso.org with JS disabled
It would be great if [haskell-miso.org](https://haskell-miso.org) worked the same with JS disabled. Which would also be a great showcase for the _isomorphic_ feature.

It seems that the only thing not functioning correctly with JS disabled are the top navigation controls. Adding a `href` attribute and `onclick="return false;"` solves this, but maybe there is a more idiomatic way to do this?

rihardsk:
Closes #241


dmjio:
Merged in https://github.com/dmjio/miso/commit/1ccb202f6c308ec900d123cb4a7fbea3fbfb3521

PR #313, opened 2017-11-09
WebSockets: refactor for error handling, no global mutable variables, etc

adnelson:
This is a refactor of the websockets subscription module.

Changes:

* Remove global mutable variables and unsafePerformIO, which follows best practices and allows for more than one websocket in an app
* Enable error handling, allowing the user to e.g. respond to a situation where the websocket fails to connect, or a message fails to parse
* Allow the reconnection behavior to be configurable by having the user pass in a function which can decide whether to reconnect
* Enable a simple ping behavior which can be useful to prevent the browser from closing an inactive socket
* Prevent invalid behavior such as closing a websocket before it has opened or sending a message to a closed socket, by using MVar. Also there’s no need for a thread to poll whether the socket has closed
* Allow apps to send plain text as well as ToJSON objects

Despite the large list above, the external interface has only changed a bit. As can be seen in the updated websockets example, there are relatively few things that need to be changed in current code. The main change is that the user is responsible for creating the socket object, which must also be used when sending data.

### DIFF
--- a/ghcjs-src/Miso/FFI.hs
+++ b/ghcjs-src/Miso/FFI.hs
@@ -20,6 +20,9 @@ module Miso.FFI
    , consoleLog
    , stringify
    , parse
+   , parseString
+   , safeParse
+   , safeParseString
    , copyDOMIntoVTree
    , item
    , jsvalToValue
@@ -101,7 +104,8 @@ foreign import javascript unsafe "console.log($1);"
 foreign import javascript unsafe "$r = JSON.stringify($1);"
   stringify' :: JSVal -> IO JSString
 
-foreign import javascript unsafe "$r = JSON.parse($1);"
+-- | Parse a raw string into JSON. Accepts any JS value.
+foreign import javascript unsafe "$r = JSON.parse('' + $1);"
   parse' :: JSVal -> IO JSVal
 
 -- | Converts a JS object into a JSON string
@@ -109,15 +113,32 @@ stringify :: ToJSON json => json -> IO JSString
 {-# INLINE stringify #-}
 stringify j = stringify' =<< toJSVal (toJSON j)
 
--- | Parses a JSString
+-- | Parses a JSVal (which of course should be a string). Less safe
+-- | than 'parseString'.
+parseString :: FromJSON json => JSString -> IO json
+{-# INLINE parseString #-}
+parseString = parse . unsafeCoerce
+
+-- | Parses a JSVal (which of course should be a string). Less safe
+-- | than 'parseString'.
 parse :: FromJSON json => JSVal -> IO json
 {-# INLINE parse #-}
-parse jval = do
-  k <- parse' jval
-  Just val <- jsvalToValue k
-  case fromJSON val of
-    Success x -> pure x
-    Error y -> error y
+parse jval = safeParse jval >>= \case
+  Success obj -> pure obj
+  Error e -> error ("When parsing JSON: " ++ e)
+
+-- | Parse a JSVal into a result, so that it can be checked.
+safeParse :: FromJSON obj => JSVal -> IO (Result obj)
+{-# INLINE safeParse #-}
+safeParse jval = do
+  parsed <- parse' jval
+  Just val <- jsvalToValue parsed
+  pure $ fromJSON val
+
+-- | Parse a JSString into a result, so that it can be checked.
+safeParseString :: FromJSON obj => JSString -> IO (Result obj)
+{-# INLINE safeParseString #-}
+safeParseString = safeParse . unsafeCoerce
 
 -- | Indexing into a JS object
 foreign import javascript unsafe "$r = $1[$2];"
--- a/ghcjs-src/Miso/FFI.hs
+++ b/ghcjs-src/Miso/FFI.hs
@@ -19,10 +19,11 @@ module Miso.FFI
    , now
    , consoleLog
    , stringify
-   , parse
-   , parseString
-   , safeParse
-   , safeParseString
+   , parseJSValAsJson
+   , unsafeParseJSValAsJson
+   , unsafeParseJsonString
+   , parseJsonString
+   , parseJsonToJSVal
    , copyDOMIntoVTree
    , item
    , jsvalToValue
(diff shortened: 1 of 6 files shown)


ghcjs-src/Miso/FFI.hs
```diff
@@ -101,23 +104,41 @@ foreign import javascript unsafe "console.log($1);"
...
 stringify :: ToJSON json => json -> IO JSString
 {-# INLINE stringify #-}
 stringify j = stringify' =<< toJSVal (toJSON j)
 
--- | Parses a JSString
+-- | Parses a JSVal (which of course should be a string). Less safe
+-- | than 'parseString'.
+parseString :: FromJSON json => JSString -> IO json
+{-# INLINE parseString #-}
+parseString = parse . unsafeCoerce
```
cocreature: No need for `unsafeCoerce`, you can use `jsval` here.

ghcjs-src/Miso/FFI.hs
```diff
@@ -101,23 +104,41 @@ foreign import javascript unsafe "console.log($1);"
 foreign import javascript unsafe "$r = JSON.stringify($1);"
   stringify' :: JSVal -> IO JSString
 
-foreign import javascript unsafe "$r = JSON.parse($1);"
+-- | Parse a raw string into JSON. Accepts any JS value.
+foreign import javascript unsafe "$r = JSON.parse('' + $1);"
```
cocreature: This seems fishy, when is this needed?

ghcjs-src/Miso/FFI.hs
```diff
@@ -101,23 +104,41 @@ foreign import javascript unsafe "console.log($1);"
...
+{-# INLINE safeParse #-}
+safeParse jval = do
+  parsed <- parse' jval
+  Just val <- jsvalToValue parsed
+  pure $ fromJSON val
+
+-- | Parse a JSString into a result, so that it can be checked.
+safeParseString :: FromJSON obj => JSString -> IO (Result obj)
+{-# INLINE safeParseString #-}
+safeParseString = safeParse . unsafeCoerce
```
cocreature: Same here.

ghcjs-src/Miso/FFI.hs
```diff
@@ -101,23 +104,41 @@ foreign import javascript unsafe "console.log($1);"
...
 
--- | Parses a JSString
+-- | Parses a JSVal (which of course should be a string). Less safe
+-- | than 'parseString'.
+parseString :: FromJSON json => JSString -> IO json
+{-# INLINE parseString #-}
+parseString = parse . unsafeCoerce
+
+-- | Parses a JSVal (which of course should be a string). Less safe
+-- | than 'parseString'.
```
cocreature: This should reference to `safeParseString`. I would also suggest to remove the `safe` prefix and instead add an `unsafe` prefix to this function. The docs should also mention in what way it is unsafe.

ghcjs-src/Miso/Subscription/WebSocket.hs
```diff
@@ -151,48 +129,207 @@ data SocketState
   | CLOSED     -- ^ 3
   deriving (Show, Eq, Ord, Enum)
 
--- | Retrieves current status of `WebSocket`
-getSocketState :: IO SocketState
-getSocketState = do
-  Just ws <- readIORef websocket
-  toEnum <$> getSocketState' ws
+-- | Low-level websocket handle, wraps the raw JSVal.
+newtype WebSocket_ = WebSocket_ JSVal
```
cocreature: Maybe call this `JSWebSocket` to make it easier to infer what this represents based on the name.

ghcjs-src/Miso/Subscription/WebSocket.hs
```diff
@@ -151,48 +129,207 @@ data SocketState
...
+  :: FromJSON message
+  => WebSocketConfig                    -- ^ Configuration
+  -> WebSocket                          -- ^ Handle to the socket.
+  -> (WebSocketEvent message -> action) -- ^ Event handler.
+  -> Sub action model                   -- ^ A subscription
+websocketSubWithConfig config socket handler getModel sink = do
+  let (url, protocols) = (wsURL socket, wsProtocols socket)
+
+  -- Spin off ping thread if configured
+  pingThreadId <- case websocketPing config of
```
cocreature: You can simplify this using `for`.

ghcjs-src/Miso/Subscription/WebSocket.hs
```diff
@@ -151,48 +129,207 @@ data SocketState
...
+        rawData -> parseData rawData
+      Nothing -> getData >=> parseData
+
+  -- When the socket closed, determine whether to reconnect. If so,
+  -- update the MVar in the WebSocket. If reconnection fails, the
+  -- WebSocketConnectFailed event will be fired.
+  onClose socket_ =<< do
+    asyncCallback1 $ \e -> do
+      -- Stop the ping thread, if it's running.
+      maybe (pure ()) killThread pingThreadId
```
cocreature: `traverse_ killThread pingThreadId`

ghcjs-src/Miso/Subscription/WebSocket.hs
```diff
@@ -151,48 +129,207 @@ data SocketState
...
-getSocketState :: IO SocketState
-getSocketState = do
-  Just ws <- readIORef websocket
-  toEnum <$> getSocketState' ws
+-- | Low-level websocket handle, wraps the raw JSVal.
+newtype WebSocket_ = WebSocket_ JSVal
+
+-- | Create a new low-level websocket.
+newWebSocket_ :: MisoString -> [MisoString] -> IO (Either MisoString WebSocket_)
+{-# INLINE newWebSocket_ #-}
```
cocreature: I’m a bit suspicious of all these `INLINE` pragmas. They tend to be a good way to ruin compile times.

ghcjs-src/Miso/FFI.hs
adnelson: So the problem there is that `JSON.parse()` expects a string, but the type signature for `parse'` accepts any `JSVal`. So you can get obscure exceptions thrown.

Admittedly the better solution would be to refactor the code so that `parse'` takes ` JSString` instead, and/or to wrap the `JSON.parse()` in a `try`/`catch` so that we can handle a JSON parse exception. I was actually planning on doing that but didn't get there because I didn't want this PR to get too large. I could certainly revert this change, but on the other hand, it doesn't make the code any *less* safe than it already is.

ghcjs-src/Miso/FFI.hs
adnelson: Didn't know about that one, thanks!

ghcjs-src/Miso/FFI.hs
adnelson: Totally in favor of that; I just didn't want to rename a function which is used throughout the code. Can do though.

ghcjs-src/Miso/Subscription/WebSocket.hs
adnelson: Sounds good

ghcjs-src/Miso/Subscription/WebSocket.hs
adnelson: The original had these on many of the functions so I was just following the convention. I'm totally fine with removing these.

ghcjs-src/Miso/Subscription/WebSocket.hs
adnelson: simpler, but less readable? I guess I'm fine with it but sometimes using combinators obscures things.

ghcjs-src/Miso/Subscription/WebSocket.hs
adnelson: works for me

adnelson: Actually I'd kind of prefer using `mapM_` since it's part of the prelude

adnelson:
@cocreature thanks for taking a look at this!

* I renamed the JSON parse functions, and made the ones that I had written safer (by catching `JSON.parse()` failures). This added some SLOC and possibly would be better in a separate PR, but I just figured I'd add it here.
* Renamed `WebSocket_` -> `JSWebSocket`
* Use `mapM` and `mapM_` (if you feel strongly about `for` or `traverse_`, I can use them)
* Since the lower-level parse functions now take `JSString` instead of `JSVal`, I still have to use `unsafeCoerce` :( but, sooner or later those functions might be able to be gotten rid of anyway and only the safe `JSString` functions used.
* Removed some of the `INLINE` pragmas but kept others. I really don't understand the tradeoffs well enough, open to inlining whatever, or not.

PR #315, opened 2017-11-16
Create AppContext type, allow to be initialized separateely

adnelson:
This maintains the existing API; in particular the current `miso` and `startApp` functions maintain the same behavior as they had before, but allows users with particular needs access to a limited portion of the internals of miso operation.

This puts the `writeEvent`, `notify` and `modelRef` objects into an `AppContext` type rather than hiding them in the internals of the `miso` and `startApp` functions. The user can initialize the `AppContext` themselves (with `newAppContext`) and use it to e.g. start or stop subscriptions during an IO action returned in an `Effect`, rather than having a fixed set of subscriptions which are opaque to the user. The ability to do this was essential for an app I'm working on.

In addition to the changes needed to support the above, there is a bit of code rearrangement in the `Miso` module.

### DIFF
--- a/ghcjs-src/Miso.hs
+++ b/ghcjs-src/Miso.hs
@@ -15,7 +15,9 @@
 ----------------------------------------------------------------------------
 module Miso
   ( miso
+  , misoWithContext
   , startApp
+  , startAppWithContext
   , module Miso.Effect
   , module Miso.Event
   , module Miso.Html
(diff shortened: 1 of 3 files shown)


cocreature:
fwiw you can do something very similar without changing anything in miso:

* Create a new channel of actions.
* Make a single subscription that reads from this channel. This subscription is never stopped.
* Now you can just create new threads that write to this channel and start/stop them as needed.

I’m not saying that’s necessarily a better way to handle this but it’s an alternative point in the design space that’s worth considering.

adnelson:
closing due to #317

PR #321, opened 2017-12-01
Add fromMisoString to `ToMisoString` typeclass

cocreature:
This can be useful when an event provides a `MisoString` but you want
to pass that `MisoString` to an API that expects `Text`.


dmjio:
@cocreature thanks 🍺

PR #332, opened 2018-01-04
Add pre-rendering attribute exceptions

dmjio:


tolgap:
@dmjio I noticed that `hidden_` is inside that list of exceptions, but it's implemented as a `textProp` 😄 .

https://github.com/dmjio/miso/blob/master/src/Miso/Html/Property.hs#L154

Might as well change this to `boolProp` in this PR 👍 .

dmjio:
yea, `selected_` as well.

dmjio:
LGTM 👍

PR #336, opened 2018-01-11
Pin nixpkgs, update README

dmjio:


dmjio:
This temporarily fixes the build and allows things to "just work" for new users.

PR #339, opened 2018-01-13
Single build of `Mario` example

sectore:
As discussed in https://haskell-miso.slack.com/archives/C2GPQ2XH7/p1515516892000400 it would be nice to build an example separately and not all at once. 

This PR enables a single build of  `Mario` example w/o the need of building all other examples. It includes a tiny code fix as well - just cosmetic :icecream: .

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,6 +444,8 @@ cd result/examples/todo-mvc.jsexe && python -m SimpleHTTPServer
 Serving HTTP on 0.0.0.0 port 8000 ...
 ```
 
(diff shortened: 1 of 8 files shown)


dmjio:
I'm not opposed to this idea, it would mean adding a lot more cabal files and nix scripts. Maybe we could get @cleverca22 to do an audit of the build process. I just don't want to hit a scenario where we have to pin the version of miso in the same repo, this could also introduce stack / nix dep. drift, but I don't think it should be too much of an issue otherwise.

dmjio:
Keeping it all in one cabal file is fine for now. Will revisit this / potentially re-open as the number of executable sections become too large.

PR #342, opened 2018-01-13
Add Util for miso-ghcjs

dmjio:
Per @osa1's recommendation


dmjio:
@osa1 is there anything else you'd like to add before we merge?

osa1:
@dmjio I need `jsvalToValue`, `consoleLog` and `now` so this covers everything I need. Thanks!

PR #352, opened 2018-02-01
Fix "disabled" attribute

basvandijk:
The "disabled" attribute should not be set to the string "false" when the value is False. Instead we shouldn't set it at all. See:

https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/disabled

I'm not sure this is the best fix but I wrote it in a hurry to reach a deadline and it worked correctly.

### DIFF
--- a/ghc-src/Miso/Html/Internal.hs
+++ b/ghc-src/Miso/Html/Internal.hs
@@ -35,6 +35,7 @@ module Miso.Html.Internal (
   , NS     (..)
   -- * Setting properties on virtual DOM nodes
   , prop
+  , optionalProp
   -- * Setting CSS
   , style_
   -- * Handling events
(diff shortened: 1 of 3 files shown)


FPtje:
This problem turns out to have arisen from a combination of the CSS framework (bulma) and using `a_` instead of `button_`. Take the following example:

```haskell
foo m = a_ [ class_ "button is-success", disabled_ $ m ^. someBool ] [text "yes"]
```
Now imagine `someBool` in the `Model` is `True`. Since `<a>` does not support the Javascript property `disabled`, Miso decides to set an attribute called `disabled` to the string `"True"`. Bulma picks this up and renders the `<a>` element as disabled. You can still click it, though, and any `onClick` listeners *will* fire.

Things will work if this code is used:
```haskell
foo m = button_ [ class_ "button is-success", disabled_ $ m ^. someBool ] [text "yes"]
```

The element will look exactly the same. Note that the `button` class *still* needs to be set. Since the `<button>` element supports the `disabled` property, `disabled_` works correctly.

Thanks @dmjio for helping us figure this out.

basvandijk:
Ok I was assuming that attributes were always set using the `setAttribute` method of a node. [According to the `setAttribute` documentation](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) "Boolean attributes are considered to be true if they're present on the element at all, regardless of their actual value".

After reading [this section of the `diffProps` function](https://github.com/dmjio/miso/blob/master/jsbits/diff.js#L98:L101) I see that we only set attributes using the `setAttribute` function if the attribute is not already present as a property of the node. Apparently `disabled` is always an existing property of a new node meaning that, instead of using `setAttribute`, we set it using `node[n] = ...` which actually does The Right Thing.

Closing, sorry for the noise.

PR #354, opened 2018-02-10
ToMisoString instances for Float, Double and Int

FPtje:
Creates instances of `ToMisoString` for `Float`, `Double` and `Int`.  Also updates the examples to reflect the change.

Some notes:

- `sample-app/default.nix` will need to pin Miso to a commit on or after the one that merges this PR, otherwise it won't compile
- Same for the `README.md` (line 268)
- The `fromMisoString` function assumes the type of the js object and that it can be parsed as a number. It means the function is unsafe. Not *too* bad if the end user knows what they're doing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,12 +412,12 @@ main :: IO ()
 main = startApp App {..}
   where
     initialAction = SayHelloWorld -- initial action to be executed on application load
-    model  = 0                    -- initial model
-    update = updateModel          -- update function
-    view   = viewModel            -- view function
-    events = defaultEvents        -- default delegated events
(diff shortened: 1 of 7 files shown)


FPtje:
We could also go to `String` and `Read` it, or `Text` and parse it some other way

dmjio:
Great work :) last request, might be beneficial to update the README.md as well, to reflect the changes to sample-app.

FPtje:
Updated README with changes in the sample application. Note that the pinned version of Miso in the README needs to change to the latest revision too after this PR has been merged.

PR #357, opened 2018-02-12
Updated Miso's hashes

FPtje:
Fixes temporary build failure regression


FPtje:
Couldn't do this earlier because I didn't know the revision numbers :D

Issue #359, opened 2018-02-16
Miso routing fails to work with url fragments.


pamu:
### Thanks for the Cool frontend framework! Miso is new hope in Haskell frontend world.

In the example https://github.com/dmjio/miso/blob/master/examples/router/Main.hs

```haskell
-- | Type-level routes
type API   = About :<|> Home
type Home  = View Action
type About = "#about" :> View Action  -- introduced # in front of about
```

I want router to work with fragments also because, this way I can use urls without invoking the backend actions in case of same domain name for frontend and backend.

Using `#` client-side routing is much more effective and flexible. As server ignores fragments. When client-side url (url with `#`) is changed browser does not request the server.

For example:
Let's say current URL is `http://foobar.com/#locations/all` now let's change the URL to `http://foobar.com/#locations/1234` by hand. Now Browser page is not refreshed. The client app can read the fragment information to load necessary data without page reloading.

pamu:
Found the solution:  https://github.com/dmjio/miso/pull/360

dmjio:
cc @Archaeron is running into a similar issue, this patch seems like the fix. Great work @pamu !

pamu:
@dmjio Shall I go ahead and update the examples/demos apps as well.

pamu:
@dmjio I am currently using Miso with this fix for my frontend app. Waiting for the fix to be merged.

dmjio:
https://github.com/dmjio/miso/pull/360#issuecomment-377201301

pamu:
@archaeron are you still interested in this ?

archaeron:
I still believe this is useful, but I don't use miso right now

dmjio:
Re-open if necessary please.

pamu:
Properly fixing this would require changes in servant repo

PR #360, opened 2018-02-16
Fix for "Miso routing fails to work with url fragments."

pamu:
# Fix for the issue  https://github.com/dmjio/miso/issues/359
Support for fragments based routing. Client URLs can be bookmarked and opened by entering the URL directly into the browser. Client routing URLs doesn't invoke server routes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,8 +265,8 @@ let
   result = import (pkgs.fetchFromGitHub {
     owner = "dmjio";
     repo = "miso";
-    sha256 = "03w1gffzw0qry6q5may5miyjf36574kx12pviywjddpckp62p8h3";
-    rev = "2a42c67ef1f876b2f81dbcf18bcc18269d0f6876";
+    sha256 = "1l1gwzzqlvvcmg70jjrwc5ijv1vb6y5ljqkh7rxxq7hkyxpjyx9q";
+    rev = "95f6bc9b1ae6230b110358a82b6a573806f272c2";
   }) {};
 in pkgs.haskell.packages.ghcjs.callPackage ./app.nix {
   miso = result.miso-ghcjs;
--- a/sample-app/default.nix
+++ b/sample-app/default.nix
@@ -9,8 +9,8 @@ let
   result = import (pkgs.fetchFromGitHub {
     owner = "dmjio";
     repo = "miso";
-    sha256 = "03w1gffzw0qry6q5may5miyjf36574kx12pviywjddpckp62p8h3";
-    rev = "2a42c67ef1f876b2f81dbcf18bcc18269d0f6876";
+    sha256 = "1l1gwzzqlvvcmg70jjrwc5ijv1vb6y5ljqkh7rxxq7hkyxpjyx9q";
+    rev = "95f6bc9b1ae6230b110358a82b6a573806f272c2";
   }) {};
 in pkgs.haskell.packages.ghcjs.callPackage ./app.nix {
   miso = result.miso-ghcjs;


dmjio:
I think we need to make `Hash` a first-class combinator, and remove it during `ToServerRoutes`, since `Hash` doesn't make sense on the server (hashes aren't transferred in HTTP requests). This also absolves us from adding it to `servant-server`, `servant-client`, etc. Also, there are rules where the hash should exist `.com/#/foo`, `.com/#bar`, `.com#/bar` or `.com#foo` etc.

dmjio:
Define
```haskell
data Hash (a :: k)
```

Remove it so server doesn't have to deal with it
```haskell
-- | Convert client route type to a server web handler type
type family ToServerRoutes (layout :: k) (wrapper :: * -> *) (action :: *) :: k where
  ToServerRoutes (Hash a :> b) wrapper action = ToServerRoutes b wrapper action
  -- ... extra cases go here
```

We might also need to account for it in `HasRouter`. https://github.com/dmjio/miso/blob/master/src/Miso/Router.hs#L192-L196

That would entail extending `Location` to account for hashes, we'd need to use `uriFragment` and handle the fragments after the paths have been dealt with iiuc.

pamu:
Will create new pull request with all code pushed into a separate branch.

RobertFischer:
Ooooooooooooooooooooooooooooh.  Please ping me in the new pull request, @pamu. Very excited about this.

pamu:
@dmjio 

For generating `safeLink` when `Hash` is used `HasLink` instance must be provided, which cannot be provided unless `Hash` is part of Servant. 

```haskell

module Servant.Utils.Links (
  module Servant.API.TypeLevel,

  -- * Building and using safe links
  --
  -- | Note that 'URI' is from the "Network.URI" module in the @network-uri@ package.
    safeLink
  , allLinks
  , URI(..)
  -- * Adding custom types
  , HasLink(..)
  , Link  ---  Notice Link constructors are not exposed.
  , linkURI
  , linkURI'
  , LinkArrayElementStyle (..)
  -- ** Link accessors
  , Param (..)
  , linkSegments
  , linkQueryParams
) where
```

So pull request to implement `Hash` on Servant client repo is a viable solution than implementing `Hash` in Miso.

For now we can go with 

```Haskell

uriToLocation :: URI -> Location
uriToLocation uri = Location
  { locPath = decodePathSegments $ BS.pack (uriPath uri ++ uriFragment uri)
  , locQuery = parseQuery $ BS.pack (uriQuery uri)
  }

```

pamu:
new pull request. https://github.com/dmjio/miso/pull/399

pamu:
@dmjio Maybe we could copy `Servant.Utils.Links` and override in `Miso` to incorporate `Hash`. This way we can provide `HasLink` instance.

dmjio:
> HasLink instance must be provided, which cannot be provided unless Hash is part of Servant.

Hmm, what's keeping us from making `instance HasLink Hash` where we define `Hash`?

pamu:
@dmjio Functions to change the link are hidden. And constructors of `Link` are also hidden.

pamu:
@dmjio Also in this implementation . Where is `x` coming from. I think `x` is not required.  
```
  instance (HasRouter m sublayout, FromHttpApiData x) =>
    HasRouter m (Hash sym :> sublayout) where
    type RouteT m (Hash sym :> sublayout) a = x -> RouteT m sublayout a
    route _ a f m = RHash (\x -> route (Proxy :: Proxy sublayout) a (f x) m)
```

dmjio:
Ah, then it seems there might be a case for this to exist inside of `servant`, we'd need to explicitly document this is used for frontend routing. As long as `ToServerRoutes` is in use instances for `HasServer` can be omitted.

PR #365, opened 2018-02-21
Use window.location.href for getCurrentURI

RobertFischer:
The current implementation of `getCurrentURI`/`getURI` did not return a host or port as a part of the URI, which made it awkward for constructing URLs. This fixes that by using `window.location.href` instead and feeding it into `network-uri` to parse out.


dmjio:
https://github.com/dmjio/miso/commit/f2dc6f49a4855c424c6d57068bd4b4be6d1558d5 🥂

PR #366, opened 2018-02-22
delegate.js speedup effort

RobertFischer:
Piggybacking off of https://github.com/dmjio/miso/pull/298 -- took a swing at implementing a more efficient `delegate.js`.

I stayed close to the algorithm currently used by `master`, which seemed mostly reasonable. The specific improvements were around the use of strict mode, `const` vs. `var`, `Array.prototype.forEach` instead of loops, and generally being pretty aggressive about factoring out common variables.

I'm unclear what the testing protocol is to test correctness and performance.


RobertFischer:
Closed it out; found a bug.

PR #368, opened 2018-02-23
Moved withFoldable and conditionalViews

FPtje:
I made the mistake in #355 to leave out the ghc side of Miso in the helper functions. I've moved them over to a new module `Miso.Helpers`, which is shared by both `ghc` and `ghcjs`. 

Since `Miso.Util` still exports the functions, it's backwards compatible.

Tests: I compiled and ran my isomorphic example, with one of the helpers shoehorned in somewhere to make sure that it works on both ghc and ghcjs.

### DIFF
--- a/ghc-src/Miso.hs
+++ b/ghc-src/Miso.hs
@@ -17,9 +17,11 @@ module Miso
   , module Miso.Html
   , module Miso.Router
   , module Miso.TypeLevel
+  , module Miso.Util
   ) where
 
 import           Miso.Event
 import           Miso.Html
 import           Miso.Router
 import           Miso.TypeLevel
+import           Miso.Util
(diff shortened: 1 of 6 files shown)


dmjio:
@FPtje why change it to `Miso.Helpers`? Why not keep it `Miso.Util`, and as you say, re-export from `miso/src`, this way it will show up in both `ghc` and `ghcjs`

FPtje:
Yeah, having both `Helpers` and `Util` is a bad idea. Is the above commit what you mean? It moves `Miso.Util` to `src/`. Since `Miso.Util` re-exported some functions from `Miso.FFI`, I had to take those out and re-export it in `ghcjs-src/Miso.hs` instead. Note that I'm re-exporting the `Miso.FFI` in its entirety now. `Miso.Util` only exported a subset of the functions. Is that OK?

dmjio:
I'd prefer not to expose `delegateEvent` and `copyDOMIntoVTree`, we could remedy this by putting `copyDOMIntoVTree` in `ghcjs-src/Miso.hs` and `delegateEvent` into `ghcjs-src/Miso/Delegate.hs`.

FPtje:
Done, and rebased on master

PR #369, opened 2018-03-02
Allow scheduled IO computations access to a sink

basvandijk:
As mentioned in https://github.com/dmjio/miso/pull/362#issuecomment-369561451 this slight generalization of `Effect` gives users the power to dispatch actions to the `update` function at their own discretion.

The most pressing use-case is scheduling an IO computation which creates a 3rd-party JS widget which has an associated callback. The callback can then call the sink to turn events into actions. To do this
without accessing a sink requires going via a subscription which introduces a leaky-abstraction. [Here's an example](https://github.com/basvandijk/miso-jswidget-example/commit/421b63f12fd9c14e5fc0ed588b265e4b08ae9b79).

Speaking of subscriptions, this generalization of `Effect` can actually supersede subscriptions. For example we can replace:

```haskell
mouseSub :: ((Int,Int) -> action) -> Sub action model
mouseSub f _ = \sink -> do
  windowAddEventListener "mousemove" =<< do
    asyncCallback1 $ \mouseEvent -> do
      Just x <- fromJSVal =<< getProp "clientX" (Object mouseEvent)
      Just y <- fromJSVal =<< getProp "clientY" (Object mouseEvent)
      sink $ f (x,y)
```

with:

```haskell
mouseSub :: ((Int,Int) -> action) -> Transition action model ()
mouseSub f = scheduleIOWithSink $ \sink -> do
  windowAddEventListener "mousemove" =<< do
    asyncCallback1 $ \mouseEvent -> do
      Just x <- fromJSVal =<< getProp "clientX" (Object mouseEvent)
      Just y <- fromJSVal =<< getProp "clientY" (Object mouseEvent)
      sink $ f (x,y)
```

But lets leave deprecating subscriptions for another day...

Credit to @FPtje for coming up with this generalization.

### DIFF
--- a/ghcjs-src/Miso.hs
+++ b/ghcjs-src/Miso.hs
@@ -51,7 +51,7 @@ common
   :: Eq model
   => App model action
   -> model
-  -> ((action -> IO ()) -> IO (IORef VTree))
+  -> (Sink action -> IO (IORef VTree))
   -> IO b
 common App {..} m getView = do
   -- init Notifier
@@ -123,7 +123,7 @@ startApp app@App {..} =
 
 -- | Helper
 foldEffects
-  :: (action -> IO ())
+  :: Sink action
   -> (action -> model -> Effect action model)
   -> (model, IO ()) -> action -> (model, IO ())
 foldEffects sink update = \(!model, !as) action ->
@@ -132,4 +132,4 @@ foldEffects sink update = \(!model, !as) action ->
       where
         newAs = as >> do
           forM_ effs $ \eff ->
-            void $ forkIO (sink =<< eff)
+            void $ forkIO (eff sink)
--- a/ghcjs-src/Miso/Effect.hs
+++ b/ghcjs-src/Miso/Effect.hs
@@ -11,10 +11,11 @@ module Miso.Effect (
   module Miso.Effect.Storage
 , module Miso.Effect.XHR
 , module Miso.Effect.DOM
-, Effect (..)
+, Effect (..), Sink
 , noEff
 , (<#)
 , (#>)
+, effectWithSink
 ) where
 
 import Data.Bifunctor
@@ -25,11 +26,15 @@ import Miso.Effect.DOM
 
 -- | An effect represents the results of an update action.
 --
--- It consists of the updated model and a list of actions. Each action
--- is run in a new thread so there is no risk of accidentally
--- blocking the application.
+-- It consists of the updated model and a list of IO computations. Each IO
+-- computation is run in a new thread so there is no risk of accidentally
+-- blocking the application. The IO computation is given a 'Sink' callback which
+-- can be used to dispatch actions which are fed back to the @update@ function
 data Effect action model
-  = Effect model [IO action]
+  = Effect model [Sink action -> IO ()]
+
+-- | Function to asynchronously dispatch actions to the 'update' function.
+type Sink action = action -> IO ()
 
 instance Functor (Effect action) where
   fmap f (Effect m acts) = Effect (f m) acts
@@ -45,7 +50,7 @@ instance Monad (Effect action) where
       Effect m' acts' -> Effect m' (acts ++ acts')
 
 instance Bifunctor Effect where
-  bimap f g (Effect m acts) = Effect (g m) (map (fmap f) acts)
+  bimap f g (Effect m acts) = Effect (g m) (map (\act -> \sink -> act (sink . f)) acts)
 
 -- | Smart constructor for an 'Effect' with no actions.
 noEff :: model -> Effect action model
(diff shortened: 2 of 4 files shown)


basvandijk:
Replacing subscriptions with transitions might look like this: https://github.com/LumiGuide/miso/commit/5fcaf9c19353c17c0e3852b222b13605d566fc5b.

The nice thing about having subscriptions as transitions is that you don't necessarily have to start them at initialization time. For example you can start a websocket subscription after the user has pressed a button. This allows more flexible and powerful applications.

dmjio:
Am very favorable towards this change, for the following reasons:

- After @FPtje discovered how to construct pure components, it was never clear how `Sub` would compose under this architecture. Introducing `Effect` as `Effect model [Sink action -> IO ()]`, not only obviates `Sub`, it allows dynamic addition / removal of subscriptions and plays nicely with @FPtje's `lifecycle-events`, allowing pure components to integrate with third party js libs very cleanly. It seems to kill two birds with one stone.

- Adding and removing subscriptions dynamically was always something I couldn't figure out how to do cleanly (was previously constructing `IO` actions that were created in a closure and passing them in as an argument to the `update` function). This required threading through various components, was a tad ugly.

- Threading actions in a `TChan` up the chain of components works, but isn't good design, nor user-friendly IMO.

- I think this will fit very nicely with hooks `onCreated` and `onDestroyed` in @FPtje's lifecycle-events branch.

Some caveats,

- I'd like to advertise this change as something for advanced-users. In general, `sink` should be considered an internal primitive, to be treated with care.

- To accommodate new users who aren't as familiar with `lens` and `transformers`, instead of defining the examples in terms of `scheduleIOWithSink`, can we define them in terms of `effectWithSink`?

```haskell
effectWithSink :: model -> (Sink action -> IO ()) -> Effect action model
effectWithSink model = \sinkAction -> Effect model [sinkAction]
```

This is the non-monadic counterpart to `scheduleIOWithSink` and would be what we suggest to new users who want third-party lib integration, but don't fully grok `lens` / `transformers`.

- Can we define `type Sink a = a -> IO ()`  alongside `Effect`

- Instead of `mouseSub`, can we now redefine it to something like `startMouseMove`, and maybe even have an `endMouseMove`, and replicate this pattern elsewhere (i.e. `websockets`).

Overall, excited about this addition, and simplification of the API.

basvandijk:
Thanks for the comments David. I've added `effectWithSink`.

I think about how to rework the subscriptions and will do that in another commit / PR so that this can already be merged if you're happy with it.

dmjio:
Sounds good, still wondering if complete deprecation of `Sub` is the right move. Maybe a top-level `Sub` w/ access to the current model has its place. Will ponder more. 

Great work! @FPtje @basvandijk

dmjio:
@basvandijk, @FPtje had an idea.

Suppose we allowed access to the model as well inside of `effectWithSink`, more or less making it `effectWithSub`.

```haskell
effectWithSub :: model -> (IO model -> action -> IO ()) -> Effect action model
-- effectWithSub :: model -> Sub action model -> Effect action model
effectWithSub model = \subAction -> Effect model [sinkAction]
```

Then we could reuse all existing `Sub`'s with `effectWithSub`, and provide `stop/start` variants of all of them. This also preserves the functionality that @FPtje needs currently.

```haskell
update model StartMouse = effectWithSub model (startMouseSub Coords)
update model StopMouse = effectWithSub model stopMouseSub

data Action
  = StartMouse
  | StopMouse
  | Coords
  deriving (Show, Eq)

startMouseSub :: ((Int,Int) -> action) -> Sub action model
startMouseSub f = \_ sink -> do
  windowAddEventListener "mousemove" =<< do
    asyncCallback1 $ \mouseEvent -> do
      Just x <- fromJSVal =<< getProp "clientX" (Object mouseEvent)
      Just y <- fromJSVal =<< getProp "clientY" (Object mouseEvent)
      sink $ f (x,y)

stopMouseSub :: ((Int,Int) -> action) -> Sub action model
stopMouseSub f = \_ sink -> windowAddEventListener "mousemove" 
```

basvandijk:
@FPtje and I debated this idea when writing this PR. I remember the argument against this was that in the `update` function the user already has access to the model. If the user also has access to `IO model` she could confuse the two models and make mistakes.

On the other hand, having access to `IO model` does unify subscriptions with `effectWithSink` (or `effectWithSub`) and allows more flexible code. 

Another issue is that we currently don't have any subscriptions nor examples that actually make use of the `IO model` computation. My concern is that we don't have enough experience yet to decide if the current `IO model` API is the best API. For example, is it better to have the current non-blocking semantics or is it better to have a blocking `IO model` computation that blocks until the model is changed. Or maybe we need to have both.

dmjio:
> Another issue is that we currently don't have any subscriptions nor examples that actually make use of the IO model computation.

Yes, this is my worry as well. 

> My concern is that we don't have enough experience yet to decide if the current IO model API is the best API.

Agreed, I believe I added it prematurely.

> For example, is it better to have the current non-blocking semantics or is it better to have a blocking IO model computation that blocks until the model is changed.

Yes, good point. It isn't clear yet. My current solution for introducing blocking behavior is to introduce an `MVar`, depicted below.

```haskell
updateModel :: Action -> Model -> Effect Action Model
updateModel ReadFile m = m <# do
  fileReaderInput <- getElementById "fileReader"
  file <- getFile fileReaderInput
  reader <- newReader
  mvar <- newEmptyMVar
  setOnLoad reader =<< do
    asyncCallback $ do
      r <- getResult reader
      putMVar mvar r
  readText reader file
  SetContent <$> readMVar mvar
```

Wrapping this pattern in something like a `getModelOnChanged` would be premature as well imo.

PR #370, opened 2018-03-07
Initial commit of code-coverage, updates to diff logic

dmjio:
Introduces 100% code coverage on `diff.js`, uses jest. 

- [ ] add nix scripts and s3 bucket for hosting coverage details.

### DIFF
--- a/jsbits/diff.js
+++ b/jsbits/diff.js
@@ -228,9 +228,7 @@ function syncChildren(os, ns, parent, doc) {
 	   -> [ c b a ] <- new children
 	*/
 	else if (oFirst.key === nLast.key && nFirst.key === oLast.key) {
-	    node = oFirst.nextElementSibling;
-	    oLast.domRef.replaceWith(oFirst.domRef);
-	    parent.insertBefore(oLast.domRef, node);
+	    swapDomRefs(node, oFirst.domRef, oLast.domRef, parent);
 	    swap(os, oldFirstIndex, oldLastIndex);
 	    diff(os[oldFirstIndex++], ns[newFirstIndex++], parent, doc);
 	    diff(os[oldLastIndex--], ns[newLastIndex--], parent, doc);
@@ -304,9 +302,7 @@ function syncChildren(os, ns, parent, doc) {
 		/* Move item to correct position */
 		swap(os, oldFirstIndex, tmp);
   		/* Swap DOM references */
-	        tmp = oFirst.nextElementSibling;
-	        node.domRef.replaceWith(oFirst.domRef);
-	        parent.insertBefore(node.domRef, tmp);
+		swapDomRefs(tmp, oFirst.domRef, node.domRef, parent);
 		/* optionally perform `diff` here */
 		diff(os[oldFirstIndex++], nFirst, parent, doc);
 		/* increment counters */
@@ -335,10 +331,14 @@ function syncChildren(os, ns, parent, doc) {
     }
 }
 
+function swapDomRefs(tmp,a,b,p) {
+  tmp = a.nextElementSibling;
+  b.replaceWith(a.domRef);
+  p.insertBefore(b, tmp);
+}
+
 function swap(os,l,r) {
   var k = os[l];
   os[l] = os[r];
   os[r] = k;
 }
-
-module.exports = diff;
--- a/jsbits/diff.js
+++ b/jsbits/diff.js
@@ -300,9 +300,9 @@ function syncChildren(os, ns, parent, doc) {
 		*/
 	    if (found) {
 		/* Move item to correct position */
-		swap(os, oldFirstIndex, tmp);
+		os.splice(oldFirstIndex,0, os.splice(tmp,1)[0]);
   		/* Swap DOM references */
-		swapDomRefs(tmp, oFirst.domRef, node.domRef, parent);
+		parent.insertBefore(node.domRef, os[oldFirstIndex].domRef);
 		/* optionally perform `diff` here */
 		diff(os[oldFirstIndex++], nFirst, parent, doc);
 		/* increment counters */
--- a/jsbits/diff.js
+++ b/jsbits/diff.js
@@ -139,7 +139,7 @@ function diffCss(cCss, nCss, node) {
 }
 
 function hasKeys(ns, cs) {
-    return ns.length > 0 && cs.length > 0 && ns[0].key !== undefined;
+    return ns.length > 0 && cs.length > 0 && ns[0].key !== null && cs[0].key !== null;
 }
 
 function diffChildren(cs, ns, parent, doc) {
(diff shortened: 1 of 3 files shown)


dmjio:
@FPtje can I get a review?

jsbits/diff.js
```diff
@@ -70,13 +84,11 @@ function diffProps (cProps, nProps, node, isSvg) {
 	     if (isSvg || !(c in node))
 	       node.removeAttribute(c, cProps[c]);
 	     else
-	       node[c] = "";
+    	       node[c] = '';
```
FPtje: Some of these lines have both tab and space characters for indentation. It's not because of this PR, but is it intentional? It feels like some of the indentation fixes are because of different tab sizes in different editors.

tests/diff.test.js
```diff
@@ -0,0 +1,446 @@
...
+         };
+}
+
+// base case
+test('Should be null when diffing two null virtual DOMs', () => {
+  const document = new jsdom.JSDOM().window.document;
+  const body = document.body;
+  var c = null; n = null;
+  diff(c, n, body, document);
+  expect(c).toBe(null);
```
FPtje: I think it may be impossible for `c` *not* to be `null`. Regardless of how `diff` is implemented, it cannot alter the value of a variable in the scope from which it is called. With `c` and `n` being defined as `null`, there's nothing `diff` could do to change that.

tests/diff.test.js
```diff
@@ -0,0 +1,446 @@
...
+    var document = new jsdom.JSDOM().window.document;
+    var body = document.body;
+
+    // populate DOM
+    var currentNode = vnode('div', [], { 'id' : 'a'});
+    diff (null, currentNode, body, document)
+
+    var newNode = vnode('div', [], { 'id' : 'a'});
+    diff (currentNode, newNode, body, document)
+    expect(currentNode.domRef['id']).toBe('a');
```
FPtje: This may not accurately test whether diffing is skipped. What if the `diff` algorithm were to have an error where it replaces nodes even when they are exactly the same? You would end up with the same `id` field, but with a different `domRef` object. Perhaps this check is better?

```javascript
expect(currentNode.domRef).toBe(newNode.domRef);
```

tests/diff.test.js
```diff
@@ -0,0 +1,446 @@
...
+    expect(destroy).toBe(1);
+    expect(childDestroy).toBe(1);
+});
+
+test('Should recreate a DOM node when tags are the same but keys are different', () => {
+    var document = new jsdom.JSDOM().window.document;
+    var body = document.body;
+    var destroy = 0;
+    var currentNode =
+	vnode('div'
```
FPtje: Should this tab character be here?

tests/diff.test.js
dmjio: Yea, you're right, `c` and `n` cannot be altered in this scope if they're `null`. A better test here might be `expect(body.childNodes.length).toBe(0);`

tests/diff.test.js
dmjio: Agreed, will update.

dmjio:
cc @cocreature @FPtje can I get a code review here, 100% coverage of diff.js w/ keys patch warrants a `1.0` release.

FPtje:
You've written quite a few tests. They look good. I checked whether they match all the documented cases in the `syncChildren` function and they do.

I'm gonna run the jswidgets example on this patch to see if it works. I will edit this post with the results.

FPtje:
I just tested with the jswidgets example and found a bug. It's a really interesting one.

Reproduce by building this branch: https://github.com/FPtje/miso-jswidget-example/tree/keys-patch

When clicking the `toggle calendar visibility` button, some elements get duplicated. It's not the keys patch, as it already goes wrong in commit 5f1a033.

Check this minimal example:

```haskell
{-# LANGUAGE TemplateHaskell   #-}

module Main where


import Control.Lens ( (^.), (%=), makeLenses, to, zoom )
import Miso ( App(..), Transition )
import qualified Miso
import Miso.Html
import qualified Miso.String as Miso

data Model
   = Model
     { _mBool :: !Bool
     }
     deriving ( Eq )

makeLenses ''Model

data Action
  = ToggleBool
  | NoOp
  deriving (Show, Eq)

main :: IO ()
main =
    Miso.startApp App
      { initialAction = NoOp
      , model         = initialModel
      , update        = Miso.fromTransition . updateModel
      , view          = viewModel
      , events        = Miso.defaultEvents
      , subs          = []
      , mountPoint    = Nothing
      }

initialModel :: Model
initialModel =
    Model
    { _mBool = True
    }

updateModel :: Action -> Transition Action Model ()
updateModel action = case action of
    ToggleBool ->
      mBool %= not

    NoOp -> pure ()


-- Call the component's `viewModel` where you want it to be drawn
viewModel :: Model -> View Action
viewModel m =
    div_ []
    [ if m ^. mBool then
        div_ [] [ text "yes bool" ]
      else
        nodeHtmlKeyed "div" (Miso.toKey ("myKey" :: Miso.MisoString)) [] [ text "no bool, but key" ]
    , button_ [ onClick ToggleBool ] [text "toggle bool"]
    ]
```

Following error when clicking the button:
```
uncaught exception in Haskell main thread: TypeError: Cannot read property 'length' of undefined
rts.js:7192 TypeError: Cannot read property 'length' of undefined
    at callDestroyedRecursive (lib.js:1997)
    at callDestroyedRecursive (lib.js:1998)
    at diffVNodes (lib.js:2037)
    at diff (lib.js:1986)
    at diffChildren (lib.js:2107)
    at populate (lib.js:2028)
    at diffVNodes (lib.js:2033)
    at diff (lib.js:1986)
    at h$$zs (out.js:21734)
    at h$runThreadSlice (rts.js:8219)
```

It doesn't happen when the other `div` has no key.

FPtje:
Oddly it doesn't go wrong on commit 5f1a0333656c092f36f4758bcf951bd179f68c3c

cocreature:
> Oddly it doesn't go wrong on commit 5f1a033

I’m confused, are you only talking about the minimal example here? Because before you said the exact opposite:

> It's not the keys patch, as it already goes wrong in commit 5f1a033.

FPtje:
Whoops! Contradicting myself there. The minimal example above works in e6f50cad98751a35fceafe106da8227945e74925, which is the commit that adds the keys patch. It gives the error in commit 5f1a0333656c092f36f4758bcf951bd179f68c3c

cocreature:
And the widget example fails on both?

FPtje:
Huh, yes it does. Widget example fails on both.

dmjio:
@FPtje identified this as being a problem with the changes I introduced to `callDestroyRecursive`, have reverted in above commit.

dmjio:
cc @FPtje @cocreature @RobertFischer @srid @sectore @fizruk @basvandijk 🎉

PR #372, opened 2018-03-14
 Remove the 'IO model' argument of Sub 

basvandijk:
Here's a more radical idea: let's get rid of the `getModel :: IO model` action from `Sub`.

We don't have any subscriptions nor examples which use the `getModel` action so I think it's better to remove it until we have a clear use-case.

This simplification allows `Subs` to be scheduled from the `update` function using `scheduleSub` or `effectSub`.

### DIFF
--- a/ghcjs-src/Miso.hs
+++ b/ghcjs-src/Miso.hs
@@ -79,23 +79,22 @@ common App {..} m getView = do
   -- Process initial action of application
   writeEvent initialAction
   -- Program loop, blocking on SkipChan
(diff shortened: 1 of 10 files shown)


dmjio:
I like it. `getModel` seems premature. We can always re-add if someone comes up with a compelling case. Removing `modelRef` is also nice. Substituting `atomicModifyIORef` for `writeIORef` seems kosher as well. Overall a big win on API simplicity and potentially speed. 👍

Issue #358, opened 2018-02-14
Make childless html tags not take in a `[View action]` argument


tysonzero:
Currently `input_` has type `[Attribute action] -> [View action] -> View action`, but anything you put inside `[View action]` seems to just spill out to the side of the `<input>` tag. It seems like a simple solution to this is for `input_` (and all other childless elements) to have type `[Attribute action] -> View action`.

dmjio:
That is a good observation. Adding children to `<input>` isn't valid HTML, even though the DOM allows it. I think a PR along these lines (that restricts specifying children on `<input>`) would be beneficial.

dmjio:
@tysonzero care to make a PR? Otherwise, I can make one soon.

tysonzero:
@dmjio As long as there is no rush, I am happy to make a PR at some point.

dmjio:
We did it 🎉 @tysonzero

PR #375, opened 2018-03-18
Remove `[View action]` argument from void elements

tysonzero:
#358

### DIFF
--- a/examples/file-reader/Main.hs
+++ b/examples/file-reader/Main.hs
@@ -66,7 +66,7 @@ viewModel Model {..} = view
       , input_ [ id_ "fileReader"
              , type_ "file"
              , onChange (const ReadFile)
-             ] []
+             ]
       , div_ [] [ text info ]
       ]
 
--- a/examples/todo-mvc/Main.hs
+++ b/examples/todo-mvc/Main.hs
@@ -175,7 +175,7 @@ viewEntries visibility entries =
         , name_ "toggle"
         , checked_ allCompleted
         , onClick $ CheckAll (not allCompleted)
-        ] []
+        ]
       , label_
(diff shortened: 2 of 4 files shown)


dmjio:
It appears there are breakages on `https://haskell-miso.org` with this change. https://hydra.dmj.io/build/14152/nixlog/2, it seems bulma allows `br` and `hr` tags to contain children.

dmjio:
Let me try to put the children of `hr` and `br` tags as siblings, stay tuned.

dmjio:
Elm actually allows the possibility of child nodes being defined in a `br` tag, despite it not being w3c valid. http://package.elm-lang.org/packages/elm-lang/html/2.0.0/Html#br

tysonzero:
Hmm. I mean if it's not w3c valid doesn't that mean that browsers can implement it however they want (a.k.a undefined behavior)? Assuming that the behavior of such child elements can be relied upon in all browsers then I think it would probably be fine to leave in the ones that are useful. I'm not sure if `Elm` is the best thing to reference since even `img` takes in children when IMO it definitely should not.

FPtje:
I believe Elm's things are generated, or at least a copy paste job. Evan may not have put any thought in it. Judging by [this issue](https://github.com/facebook/react/issues/12396), React forbids void elements to have children. We could do the same. If, hypothetically, a similar issue is created on this repository, we could reply with "there's still `nodeHtml`".

It seems like a good change. It's a breaking change, though. The nice thing here is that we can enforce a HTML rule through the type system.

dmjio:
Added fixes for `haskell-miso.org`, rebased and merged in `6b3734074552476cd179660a79ba0ec6970aa144`

PR #377, opened 2018-03-20
RFC:  Use a TQueue for buffering actions

basvandijk:
Storing the actions into a [TQueue](http://hackage.haskell.org/package/stm-2.4.5.0/docs/Control-Concurrent-STM-TQueue.html) is a bit simpler than storing them in an `IORef (Seq action)` and pairing that with a MVar-based notify mechanism. Has this been tried / considered before?

All examples and tests build and work correctly. 

The performance implications still have to be analyzed. Do we have a benchmark suite?

cc @FPtje.


dmjio:
Maybe @luite can speak to the performance of the `stm` implementation in GHCJS, and maybe provide tips on benchmarking. Miso's current benchmarks suite is only used for virtual DOM, and doesn't reference the GHCJS RTS. I suppose we could compile criterion to js and run that, unsure if that would be accurate. There's also the chrome profiler. 

For me a good indicator is eyeballing the flatris example when you hold down the up arrow key.
https://unsoundscapes.itch.io/flatris <-- Elm's implementation
https://flatris.haskell-miso.org/ <-- Miso's

You'll notice Elm's rotates a lot faster, hard to tell what our bottleneck is though.

basvandijk:
I wrote some very preliminary benchmarking code in https://github.com/basvandijk/bench-queues. Initial results indicate that miso's queue is faster than a TQueue.

@luite note that the GHCJS RTS almost always crashes on benchmark B when running the miso queue benchmark. This is with ghcjsHEAD from nixpkgs at `release-17.09` which is using: https://github.com/ghcjs/ghcjs/commit/2b3759942fb5b2fc1a58d314d9b098d4622fa6b6. Any idea what is causing this and if this has been fixed in GHCJS-8.2?

basvandijk:
I'm closing this because a TQueue appears to be slower.

Issue #376, opened 2018-03-20
Consider getting rid of `HasUri` constraint on `miso`


tysonzero:
In order for `model` to have a `HasUri` instance it must contain a `URI`, so in order to create that model to pass into `miso` in the first place you must give it some sort of `URI`, which is just going to get immediately overridden.

It seems like it would make more sense for the `model` argument to `App` to have type `URI -> model`, instead of `model`, then we can drop the `HasURI` constraint. We could also give it type `model` with no constraint and just let users decide if they want to call `getCurrentURI` or not.

Funnily enough with the current setup I have actually been indirectly calling `getCurrentURI` twice, once to build the `model`, and once in `miso`, which I only just noticed after doing some source diving to figure out the purpose of `HasURI`.

tysonzero:
After further thought I would suggest simply giving it type `model` and allowing users to call `getCurrentURI`, as that would also allow you to do other things based on the URI such as change the other parameters to miso or get some of the model state from `IO`.

dmjio:
`HasURI` might be too heavyweight, I'd tend to agree. Also introduces lens where it might not be that necessary. It might be safe to assume that all applications will hold the `URI` somewhere on the top level of the model.

tysonzero:
I was also considering not even storing the URI in the state for my app, as all I really need is a function from `URI -> State` that gets dispatched whenever a link is pressed or the page is loaded. I don't really need to be able to convert from `State` back to a `URI`, since the `URI` is already going to be in the URL bar from the last time I clicked a link or loaded the page.

I just have to make sure that large state changes are done via setting the `URI`, and not via directly manipulating the state to one that doesn't fit well with the `URI`, which is something you have to worry about even if you do store the `URI` directly.

Would you like me to make a PR?

dmjio:
> In order for model to have a HasUri instance it must contain a URI, so in order to create that model to pass into miso in the first place you must give it some sort of URI, which is just going to get immediately overridden.

With the `miso` function yes, this assumes you'd like to share routes with the server, so we grab the `currentURI` for the user and route the correct place. Overriding isn't a problem here IMO. If using `startApp` no overriding occurs, but it also assumes you're not sharing with the server.

> It seems like it would make more sense for the model argument to App to have type URI -> model, instead of model, then we can drop the HasURI constraint. We could also give it type model with no constraint and just let users decide if they want to call getCurrentURI or not.

Not all applications will use routing, so unsure if it makes sense to impose it upon users.

[...]

PR #378, opened 2018-03-23
Remove HasURI constraint from `miso` function

tysonzero:
#376 

The constraint is solely used to overwrite the existing URI attached to the model with the current URI. However since you have to create a URI on the model in the first place this does not really improve developer experience. It also prevents you from having a model that instead of storing the URI, simply generates the initial state of the App based on the URI then forgets it.

### DIFF
--- a/examples/file-reader/Main.hs
+++ b/examples/file-reader/Main.hs
@@ -66,7 +66,7 @@ viewModel Model {..} = view
       , input_ [ id_ "fileReader"
              , type_ "file"
              , onChange (const ReadFile)
-             ] []
+             ]
       , div_ [] [ text info ]
       ]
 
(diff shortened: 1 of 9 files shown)


dmjio:
> The constraint is solely used to overwrite the existing URI attached to the model with the current URI. 

The motivation for `HasURI` was to ensure the user doesn't forget to supply the `URI` to the model when using the `miso` function, and also to set the `URI` on the model every time a new `pushState` event has occurred. A `Lens` makes this very convenient, and `HasURI` is a law-abiding class. If we remove `HasURI`, we'd then be trusting the user to place the `URI` on the model in the `miso` function, and the user would also have to provide a getter when calling `runRoute`. 

So the new signature could be 

```haskell
miso :: Eq model => URI -> App model action -> IO ()
```

and `runRoute` would become

```haskell
runRoute
  :: HasRouter m layout
  => Proxy layout
  -> RouteT m layout a
  -> (m -> URI)
  -> Either RoutingError a
```

I'm not sure it's worth it to remove `HasURI`, we even provide convenient `makeLens` (same as `Control.Lens.lens`), it seems like purely a stylistic choice. `HasURI`

dmjio:
#379

PR #379, opened 2018-03-23
Remove HasURI class completely

tysonzero:
#376 

Extends #378 

#378 already removes every use of `setURI` so `HasURI`'s only remaining purpose is so that `runRoute` can use `getURI`. This pull request changes `runRoute` to directly take in a `URI` and spit out the appropriate handler, which you can then pass the model into.

This change allows the user to not worry about lenses or `HasURI`. It also makes `miso` more flexible, and `runRoute` more flexible.

### DIFF
--- a/examples/file-reader/Main.hs
+++ b/examples/file-reader/Main.hs
@@ -66,7 +66,7 @@ viewModel Model {..} = view
       , input_ [ id_ "fileReader"
              , type_ "file"
              , onChange (const ReadFile)
-             ] []
+             ]
       , div_ [] [ text info ]
       ]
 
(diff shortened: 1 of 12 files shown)


dmjio:
I guess classy lenses might be too much, but I would like to keep the `m` type variable in `HasRouter`, since this allows the client handlers to have the signature `Model -> View action`. This is just another convenience for users. I'll rebase some things on top of your changes and merge when it's ready. Great work! 🍻

dmjio:
https://github.com/dmjio/miso/commit/f2dc6f49a4855c424c6d57068bd4b4be6d1558d5 🥂

PR #388, opened 2018-03-27
Implement ToApi, ToHandler and mapLeaves

tysonzero:
So I added some type level handler and api mapping functions that I am fairly certain will allow me to do the thing that I want. It is also completely backwards compatible and won't affect people who aren't interested in using it.

My server handler is basically just a mirror of my client handler, just with some wrapping Html, and I want to get rid of my current handler duplication.

You can define your routes with `()` on each endpoint in order to use the features I added.

One option is then to create a miso handler for it that maps each endpoint to a view, which you can wire up to a parent view via `runRoute` to build the client app. Then for the server you can simply run `mapLeaves` on your miso handler and add the necessary `pure` / `HtmlPage`

Another option is to create a handler that maps each endpoint to an app state. From there you can create the client by calling `runRoute` on this handler to get the state for the app, then pass that state onto the primary view. You can also create the server by calling `mapLeaves` on the original handler, passing each app state into the main view and then wrapping it in the desired `HtmlPage` / `pure` to create a Servant compatible handler.

One other thing to note is that at least for my app, sometimes depending on the URL I want to do a custom action on load, such as send an Ajax request for extra data. With this new setup that is pretty easy, you can have the router return both a state and an action, and then the server just renders the view without executing the action, but the client does execute the action on load. Or even the client and the server could both execute the action (so on the client side you would do it before calling `miso` so that the state syncs up.

Either way you avoid the current issue of having both `mainView m = case runRoute (Proxy @Routes) viewTree m of ... where viewTree = homeView :<|> fooView :<|> barView` AND `serverHandlers = go homeURI :<|> go fooURI :<|> go barURI where go u = pure . HtmlPage . mainView $ getInitialState u`.

Perhaps I am just setting up my app in a strange way and this issue doesn't truly exist, but miso-isomorphic-example does very much share this issue, see [here](https://github.com/FPtje/miso-isomorphic-example/blob/master/shared/Common.hs#L66) and [here](https://github.com/FPtje/miso-isomorphic-example/blob/master/server/Main.hs#L45), lines 44-64 of `server/Main.hs` are basically just mapping a URI to that same URI.

The one remaining concern I have is that the current `runRoute` might not be able to handle routing things that aren't miso views. I will most likely follow this up with another `runRoute` PR, but one that builds a convenience function on top that allows the old behavior to work just fine.


tysonzero:
You can actually avoid any duplications without this PR, as seen here: https://github.com/tysonzero/miso-example/tree/improved

PR #389, opened 2018-03-28
Make Router work on arbitrary (non Model -> View a) handlers

tysonzero:
As I mentioned in #388, I knew that I was probably going to have to change Router so that it worked on things besides `Model -> View a`. This patch implements such a thing, and I also created a convenience function `runMisoRoute` that implements the old routing.

With this patch you can route on arbitrarily typed handlers, as long as they are homogenous, you can use either `View a` or `()` for the types of the final endpoints, I decided against using OverlappingInstances or anything like that, although in principal you could use them to allow for any arbitrary API endpoint type.

It might be worth thinking more about this kind of thing, in the sense that you could maybe argue that APIs should have a dedicated `Endpoint` type constructor that takes in the type that the endpoint returns, be it a pure view or state or a monadic servant endpoint. If such a thing is implemented it would probably be best on the servant side, which admittedly makes such a thing kind of unlikely, as some work would have to be spend making sure that it isn't an overly breaking patch to servant.

If you would like I could try and give a more compelling example (perhaps in the examples directory) of how the above change can make implementing more complex isomorphic apps easier, as it's pretty much the only pain point I have (that I can think of) right now with Miso itself.

### DIFF
--- a/examples/haskell-miso.org/client/Main.hs
+++ b/examples/haskell-miso.org/client/Main.hs
@@ -20,7 +20,7 @@ main = miso $ \currentURI -> App
       events = defaultEvents
       subs = [ uriSub HandleURI ]
       viewModel m =
-        case runMisoRoute (Proxy :: Proxy ClientRoutes) handlers uri m of
+        case runRoute (Proxy :: Proxy ClientRoutes) handlers uri m of
           Left _ -> the404 m
           Right v -> v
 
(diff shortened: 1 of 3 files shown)


tysonzero:
Since there is no `HasLink` instance for `()`, I switched to `Raw` as the trivial endpoint.

dmjio:
@tysonzero yes, please provide an example, that would be help me understand the motivation for this change.

tysonzero:
For an example of how this PR can be used, see [here](https://github.com/tysonzero/miso-example/tree/dd90dea83f8583d6f09dae5343b988d30826d962).

PR #391, opened 2018-03-28
Register event related callbacks and clear them right before diffing

### LINKED ISSUE #363: Space leak with Callbacks
# How to reproduce

1. Open the [SVG example](https://svg.haskell-miso.org/)
1. Open Chrome's task manager with shift+escape (or some other browser's equivalent)
1. Move your mouse
1. Watch memory usage go up
1. Stop moving your mouse
1. See memory usage stop going up

# Cause 
[`onWithOptions`](https://github.com/dmjio/miso/blob/master/ghcjs-src/Miso/Html/Internal.hs#L257) attributes create callbacks every time the view function is called. All callbacks are stored in a variable called `h$extraRoots`. Unless you release them with the `releaseCallback` function (see https://github.com/ghcjs/ghcjs-base/blob/master/GHCJS/Foreign/Callback.hs#L46) or its `h$release` Javascript equivalent, they're gonna pile up.

One solution is to clean them up in the Javascript side of the diffing algorithm, as that algorithm knows when callbacks can be removed.

FPtje:
Fixes #363 

This is a really simple solution. When you create an event related callback, put it in the list. Right before diffing everything in the list is released. They're safe to release, since the view function recreates them.

Tested with my components example and the jswidgets example.

### DIFF
--- a/jsbits/delegate.js
+++ b/jsbits/delegate.js
@@ -1,3 +1,20 @@
+var registeredCallbacks = [];
+
+/* Callbacks in ghcjs need to be released. With this function one can register
+   callbacks that should be released right before diffing.
+*/
+function registerCallback(cb) {
+    registeredCallbacks.push(cb);
+}
+
+/* This clears the callbacks. */
+function clearCallbacks() {
+    for (var i in registeredCallbacks)
+      h$release(registeredCallbacks[i]);
+
+    registeredCallbacks = [];
+}
+
 /* event delegation algorithm */
 function delegate(mountPointElement, events, getVTree) {
     for (var event in events) {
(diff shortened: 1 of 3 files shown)


FPtje:
@basvandijk

jsbits/delegate.js
```diff
@@ -1,3 +1,18 @@
...
+   callbacks that should be released right before diffing.
+*/
+function registerCallback(cb) {
+    registeredCallbacks.push(cb);
+}
+
+/* This clears the callbacks. */
+function clearCallbacks() {
+    for (var i in registeredCallbacks)
+      h$release(registeredCallbacks[i]);
```
basvandijk: @FPtje shouldn't `registeredCallbacks` be reset to `[]` here?

FPtje: Good point

dmjio:
LGTM @FPtje @basvandijk. Great work.

PR #399, opened 2018-04-01
Add support for fragment routes.

pamu:


tysonzero:
It seems like this is a breaking change, anyone using anchors for their original purpose of jumping to sections of the page will now have those anchors sent through the router. If `/foo#bar` originally linked to `/foo` and jumped to `#bar`, it could now instead link to `/foo/bar`. On top of that `/foo#bar` will route to `/foo` on the server and `/foo/bar` on the client, causing serious desync issues.

PR #404, opened 2018-04-15
Add mapSub and mapAction

basvandijk:
`mapAction` and `mapSub` are handy when nesting components.

Maybe we should bikeshed over the names a bit because I didn't think too long about it.


dmjio:
Both sound great to me. Have considered newtyping `Sub` or turning it into
```haskell
newtype Sub a = Sub { runSub :: ContT () IO a }
  deriving (Functor)
```
So we can have `Functor` for free. 

But that would be a breaking change... could always add it later I suppose. Like how this PR doesn't introduce breaking changes. Thanks! 🍺

PR #405, opened 2018-04-15
Add scheduleIO_ :: IO () -> Transition action model ()

basvandijk:
Besides `scheduleIO_` I also use the following in my own code:

```haskell
scheduleIOMaybe :: IO (Maybe action) -> Miso.Transition action model ()
scheduleIOMaybe io = Miso.scheduleSub $ \sink -> do
                       mbAction <- io
                       for_ mbAction $ \action ->
                         sink action
```

I didn't include that in this PR because I haven't found a satisfying name yet...

### DIFF
--- a/ghcjs-src/Miso/Types.hs
+++ b/ghcjs-src/Miso/Types.hs
@@ -17,6 +17,7 @@ module Miso.Types
   , fromTransition
   , toTransition
   , scheduleIO
+  , scheduleIO_
   , scheduleSub
   ) where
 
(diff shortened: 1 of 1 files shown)


dmjio:
Awesome, thanks! Can definitely see a use case for `scheduleIOMaybe` as well. Regarding naming, if we generalize we could have:
```haskell
scheduleIOMapM_ :: Foldable f => IO (f action) -> Miso.Transition action model ()
```
or `scheduleIOFor_` / `scheduleIOForM_`, may also suffice. Let me know if you see issues with generalizing.

basvandijk:
I like the idea of generalizing!

+1 for the name `scheduleIOFor_`.

dmjio:
cc @basvandijk 49f4443b590b2524050957bb94d11af02e744e93 🍺

PR #417, opened 2018-05-15
RFC: Introduce LocatedView to make the location (URI) part of the View

basvandijk:
**This is a RFC and still WIP.**

Conceptually the location should be part of the view since it's what the user sees in her browser's address bar. Currently changing the URI means going via IO (`pushURI`). This is unfortunate since miso is about having a pure view function.

This patch changes the type of the view function to return:

```haskell
data LocatedView action = LocatedView
  { lvLocation :: Maybe URI
  , lvView :: View action
  }
```

When the view function returns a `LocatedView` with `lvLocation` set to `Just uri` and `uri` is different than the current URI, `uri` will be pushed on the history using `pushURI`.

@dmjio @FPtje the code isn't tested yet and I haven't even run it. For now I'm mainly interested in whether you think this is a good idea and if the API and implementation make sense.


basvandijk:
Considering the fact that in the future we want to put more things in `LocatedView` to make them pure (think about favicon, notifications, etc.) the name `LocatedView` is probably to narrow. 

Maybe `LocatedView` should be renamed to `View`
and `View` should be renamed to `DOM` or something similar...

basvandijk:
Note to self: since miso should now always know the current URI it should probably register an [onPopState](https://github.com/dmjio/miso/blob/master/ghcjs-src/Miso/Subscription/History.hs#L109) callback...

RobertFischer:
Usually, I parse the URI into a route, and it's not the URI that defines the view, but the route. The route then lives in my model, which the view has access to. So I don't really see the URI as "part" of the view, except in a derivative way.

Going the other way -- If I want to change the URI, then I fire an action, which calls my `goToRoute` function (which converts the route to a URI and pushes it).

So I guess I'm not seeing the use case for this -- when do you need to return a URI change from the view function?

RobertFischer:
Moving away from [TEA](https://guide.elm-lang.org/architecture/) / [Flux](https://facebook.github.io/flux/docs/in-depth-overview.html#content) actions, there is an architecture where the view returns the new model, so rendering the view then has a potentially mutating effect on the model. That kinda gives me the heebie-jeebies, though, because it no longer means that my update function contains all the possible ways my model can be updated.

RobertFischer:
I'm still unclear on why we're complicating the `view` function.

Why would you expect a function that calculates `model -> (View action, URI)`? I get that they are both derivations of the model, but wouldn't you expect two functions, `model -> View action`, and another one that goes `model -> URI`? Why do they need to be intertwined?

Changing the URI seems like a IO-y kind of thing, because that's changing the state of the browser. Updating the DOM is also an IO-y kind of thing, which is why you can't do it from within Miso except through FFI (which should always execute in IO, IMHO). The DOM is *not* a derivative of the model, and neither is the browser's URI -- however, the VDOM and an intended URI are both derivative of the model.

Finally, I also still don't see the use case. When do you know that you want to change the URI in `view`, but you didn't know when you last went through `update`? Part of the conceptual simplicity of Miso is that the state of the application -- that's both `model` and  the state of the browser/client -- only changes in response to `update` based on the `action` parameter passed into it.  So without a really compelling use case, I'd vote strongly against complicating things.

basvandijk:
Robert, thanks for your comments. I really appreciate them.

I had the view function as `model -> (View action, URI)` because I thought it might be simpler and possibly more efficient to have a single function that calculates all the derivatives. However I now realise that splitting them up into two separate functions is probably simpler.

I also suspect that we can have the advantages of my initial design (no manual scheduling of IO actions to keep the URI in sync) but keeping this outside of miso by having something like the following function: (this is just a sketch)

```haskell
mkSyncURI :: (model -> URI) -> IO (Transition action model -> Transition action model)
mkSyncURI getURI = do
    uri <- getCurrentURI
    uriRef <- newIORef uri
    onPopState =<< do
     asyncCallback $ do
       newURI <- getCurrentURI
       writeIORef uriRef newURI
    pure $ \transition -> do
      prevURI <- gets getURI -- as I write this I realise we shouldn't get the prevURI 
                             -- from the model but we should read it from uriRef instead. 
                             -- However that's IO so the only way to do that is to always 
                             -- schedule and IO action which seems bad. 
                             -- So maybe this function is not possible after all... 
      transition
      newURI <- gets getURI
      when (prevURI /= newURI) $ scheduleIO_ $ do
        writeIORef uriRef newURI
        pushURI newURI
```

intended to be used as:

```haskell
data Model = Model { _uri :: URI, ... }

makeLenses ''Model

data Action = ChangeURI URI | ...

main = do
    syncURI <- mkSyncURI (view uri)
    startApp App
      { 
        ...
      , update = fromTransition . syncURI . updateModel
        ...
      }

updateModel :: Action -> Transition Action Model
updateModel = \case
    ...
    ChangeURI newURI -> uri .= newURI  -- look ma, no IO ;-)
    ...
```

I think for now I'll close this and keep changing the URI by scheduling IO actions.

Thanks for all your comments.

PR #423, opened 2018-06-26
Conform all three copies of the sample/simple app

### LINKED ISSUE #422: Sample application = Simple example?
The Sample application in [README.md](https://github.com/dmjio/miso/blob/master/README.md) is very similar (and functionally identical?) to the [Simple example](https://github.com/dmjio/miso/blob/master/exe/Main.hs). But a diff shows many differences in the code.

I'd like to create a pull request to conform them. Any opinions on which one is better?

JamieBallingall:
This resolves #422

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,13 +420,13 @@ data Action
 main :: IO ()
 main = startApp App {..}
   where
-	initialAction = SayHelloWorld -- initial action to be executed on application load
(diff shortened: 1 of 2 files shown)


dmjio:
@JamieBallingall Thanks! 🍻

Issue #325, opened 2017-12-16
MathML


hlinander:
Tried to output mathml using `nodeHtml` to produce the child elements, however even if the dom looked correct on inspection with dev tools the browser does not render the content correctly. I suppose this is due to the VDom handling? I noticed the `_math` element existed so perhaps I am missing something.

dmjio:
@hlinander thanks for bringing this up, this will make a cool example. Firstly, can you paste the code in question that isn't being rendered properly?

hlinander:
For example, using the code from sample-app but replacing the viewModel with
```
viewModel _ = math_ [] [
  nodeHtml "msup" [] [
      nodeHtml "mi" [] [ text "x" ],
      nodeHtml "mn" [] [ text "2" ]
                     ]
    ]
```
generates a DOM of the form

`<math><msup><mi>x</mi><mn>2</mn></msup></math>`

which is correct but which does not render as it should (https://jsfiddle.net/vdumtzsx/), i.e. the exponent is not placed correctly but rather just following the base.

I have tried this in Firefox 57.0.1 and Chromium 63.

hlinander:
Probably the same issue as https://github.com/reflex-frp/reflex/issues/38,
i.e. these elements need to be created in the correct namespace. This happens automatically when loading a HTML page but not when injected via `document.createElement`, instead `document.createElementNS` is needed. Is it possible to specify the XML namespace for an element with miso?

dmjio:
All DOM elements are defined in terms of `node`
```haskell
node :: NS -> MisoString -> Maybe Key -> [Attribute m] -> [View m] -> View m
```

Right now I believe only SVG and HTML namespaces are supported. 
https://github.com/dmjio/miso/blob/master/jsbits/diff.js#L136-L138

I think it might be possible just specify 
```xml
<math xmlns="http://www.w3.org/1998/Math/MathML">
```
in `miso` as
```haskell
math_ = node HTML "math" [ prop "xmlns" "http://www.w3.org/1998/Math/MathML" ] 
```

Also, the fact Chrome does not support it, we'd have to ask if it's worth it. There are other solutions like `MathJAX` which I think provides fallbacks for Chrome. I am not as familiar with MathJAX though, but it should be possible to integrate.

dmjio:
@hlinander any updates on this?

dmjio:
closing due to inactivity

hlinander:
@dmjio: Hope it is ok to resurrect this issue :)

[...]

PR #439, opened 2018-07-29
Add MathML support

hlinander:
I finally found some time to get back to #325 which got closed due to inactivity a long time ago. This is a try for a minimal implementation of mathml support that I am currently using in a project. Thought it would be good to check in with you before I do anything more to see if this is heading in a good direction or if you would rather take a different approach.


dmjio:
@hlinander Looks great on firefox! 
screen shot 2018-07-29 at 6 58 42 pm

PR #440, opened 2018-07-31
Initial commit of empty notify

dmjio:
cc @tysonzero @cocreature

### DIFF
--- a/src/Miso/Concurrent.hs
+++ b/src/Miso/Concurrent.hs
@@ -11,6 +11,7 @@
 module Miso.Concurrent (
     Notify (..)
   , newNotify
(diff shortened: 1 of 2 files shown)


src/Miso/Concurrent.hs
```diff
@@ -11,6 +11,9 @@
 module Miso.Concurrent (
     Notify (..)
   , newNotify
+  , newEmptyNotify
+  , EventWriter (..)
```
cocreature: Where is this coming from? This doesn’t look like it will actually compile.

dmjio: Needed to rebase

PR #441, opened 2018-07-31
Support compiling the frontend API using JSaddle

cocreature:
Support for JSaddle is currently opt-in using a Cabal flag.

* If you compile with GHCJS and disable the flag, you should get the
same dependencies, performance, … as before.
* If you compile with GHC and disable the flag, you can still use miso
for server-side rendering.
* If you compile with GHC and enable the flag, you will get the same
  API as when compiling with GHCJS but the implementation uses
  JSaddle. This allows you to compile and run your code using one of
  the various backends for JSaddle, most notably `jsaddle-warp`.


dmjio:
🍜

dmjio:
Might be nice to merge `jsaddle.nix` with `default.nix` and use two version of `nixpkgs` (one for ghcjs, another for ghc w/ jsaddle).

dmjio:
@cocreature I can add that later. Great work. Will try to update the docs to reflect this change.

dmjio:
🍺

PR #454, opened 2018-08-24
Safari checkbox selection bug fix

dmjio:
For inputs of type checkbox, selections do not apply and should return
null. Safari throws a type error. This code checks for that condition
to ensure an error is not thrown. cc @tysonzero

### DIFF
--- a/jsbits/delegate.js
+++ b/jsbits/delegate.js
@@ -121,8 +121,14 @@ window.objectToJSON = function objectToJSON (at, obj) {
 
   /* If obj is a non-list-like object */
   var newObj = {};
-  for (var i in obj)
+    for (var i in obj){
+    /* bug in safari, throws TypeError if the following fields are referenced on a checkbox */
+    /* https://stackoverflow.com/a/25569117/453261 */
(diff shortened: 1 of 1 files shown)


dmjio:
cc @cocreature

tysonzero:
That fixes it!

cocreature:
This fixes the code for checkboxes but I would expect that safari behaves similarly for other input types where these fields do not apply. Given that there are only 4 types (text, search, email, password) where they do apply, it might be worth considering to skip this case if it the type is not one of those.

dmjio:
@cocreature it seems to only apply to `checkboxes` from a few tests.

```javascript
function blowUpSafari () {
  var input = document.createElement('input');
  input['type'] = "text";
  console.log('kaboom', input.selectionStart, input.selectionEnd, input.selectionDirection);
  input['type'] = "email";
  console.log('kaboom', input.selectionStart, input.selectionEnd, input.selectionDirection);
  input['type'] = "password";
  console.log('kaboom', input.selectionStart, input.selectionEnd, input.selectionDirection);
  input['type'] = "checkbox";
  console.log('kaboom', input.selectionStart, input.selectionEnd, input.selectionDirection);
}
```

PR #460, opened 2018-10-22
Add HasRouter instance for Header

tysonzero:
This purpose of this instance is pretty much solely for route/handler sharing between client and server. As the client will just always receive "Nothing" for the header whilst the server will receive the real header value, whereas before the client would refuse to compile the code at all.


dmjio:
🍺

PR #462, opened 2018-11-14
Add scripts to build miso with 8.4

dmjio:
cc @cocreature @FPtje

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,23 +1,43 @@
-{ pkgs ? import ((import <nixpkgs> {}).fetchFromGitHub {
-    owner = "NixOS";
-    repo = "nixpkgs";
-    rev = "a0aeb23";
-    sha256 = "04dgg0f2839c1kvlhc45hcksmjzr8a22q1bgfnrx71935ilxl33d";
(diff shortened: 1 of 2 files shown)


dmjio:
cc @tysonzero

dmjio:
Still suffers from https://github.com/ghcjs/ghcjs/issues/666, leaving this open until it's fixed.

dmjio:
@FPtje's patch fixes the build. https://github.com/ghcjs/ghcjs/issues/710

FPtje:
Check this out:
https://github.com/FPtje/miso-isomorphic-example/blob/93470ca431d759168facfac5234bbc5e4619e72c/default.nix

The version of nixpkgs in that file (and later versions of nixpkgs 18.09) have the right fixes for the above mentioned issues. There are still some overrides to account for https://github.com/ghcjs/ghcjs/issues/711.

It means you can remove the patch and use `ghc844` instead of `ghc843`.

dmjio:
Closing in favor of #493 (bumps to use `ghcjs863`)

PR #471, opened 2019-01-28
Enable Closure's ADVANCED_OPTIMIZATIONS

puffnfresh:


dmjio:
@puffnfresh I have issues with variable renaming here, but I think using `--externs` might be able to get around this. https://developers.google.com/closure/compiler/docs/api-tutorial3#externs

dmjio:
cc @puffnfresh https://github.com/dmjio/miso/pull/493, think this is looking right.

dmjio:
Closing in favor of #493, @puffnfresh this PR is correct, `miso` just wasn't exporting the right functions, so the closure compiler was renaming things that it shouldn't have. Respective fixes for this on #493

PR #477, opened 2019-04-17
Isomorphism: Fix consecutive text nodes causing an error

FPtje:
- Based on https://github.com/dmjio/miso/pull/309 (CC @cocreature)
- Rebased on latest (and greatest) master
- Created tests for isomorphism (within diff.test.js, can change if desired)

Test report:
2019-04-17-231143_3142x1966_scrot

Note: When running `npm test` I got the following error:
```
SecurityError: localStorage is not available for opaque origins
```

After some Googling I found you need to set `"testURL": "http://localhost/"` in a jest config file, which is then passed to the `./node_modules/.bin/jest --config jestconfig.json` command. This change is not included in this PR as it might be caused by the version of jest I'm running.

### DIFF
--- a/tests/diff.test.js
+++ b/tests/diff.test.js
@@ -759,7 +759,7 @@ test('Should copy simple nested DOM into VTree', () => {
     expect(currentNode.children[0].children[0].text).toEqual('foo');
 });
 
-test('Should throw exception about expecting text node', () => {
+test('Should fail because of expecting text node', () => {
     var document = new jsdom.JSDOM().window.document;
     var body = document.body;
     var div = document.createElement("div");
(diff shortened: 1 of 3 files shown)


tests/diff.test.js
```diff
@@ -743,3 +744,59 @@ test('Should diff keyed text nodes', () => {
...
+    var body = document.body;
+    var div = document.createElement("div");
+    body.appendChild(div);
+    var txt = document.createTextNode("foobarbaz");
+    div.appendChild(txt);
+    var currentNode = vnodeKids('div', [ vtext("foo"), vtext("bar"), vtext("baz") ]);
+    copyDOMIntoVTree(currentNode, document);
+    expect(currentNode.children[0].text).toEqual('foobarbaz');
+    expect(currentNode.children[1].text).toEqual('');
+    expect(currentNode.children[2].text).toEqual('');
```
dmjio: Shouldn’t this expect the children to each contain “foo”, “bar”, “bar”, respectively? This would be an example of the dom and vdom losing synchrony and should not be allowed.

FPtje:
Updated PR to maintain synchrony and also with more checks.

dmjio:
@FPtje would be nice if instead of throwing a `MisoException` we did a `console.warn('Could not copy DOM into virtual DOM, falling back to diff")`.

@tysonzero would be nice if you could test this branch out on your codebase just to be certain its correct.

tysonzero:
Our site seems to work fine on this branch. We don't have any consecutive `text` nodes though so it's mostly just a regression test.

PR #481, opened 2019-04-25
jsbits: Fixed indentation

FPtje:
Tabs and spaces were used interchangeably per line, and there was no setting on my editor
that dislayed it right. 

This PR only deals with the indentation. I've tried to avoid changing the formatting otherwise.


dmjio:
You can thank `js-mode` on emacs for this :)

PR #491, opened 2019-07-03
Update sample-app according to the instructions in the README

FranklinChen:
This should have been done in https://github.com/dmjio/miso/pull/485 but was omitted there.


dmjio:
Thanks @FranklinChen  !

PR #493, opened 2019-07-06
Export all functions to allow advanced optimizations.

dmjio:
https://developers.google.com/closure/compiler/docs/api-tutorial3#export

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,32 +1,53 @@
-{ pkgs ? import ((import <nixpkgs> {}).fetchFromGitHub {
+{ nixpkgs ? (import <nixpkgs> {}).fetchFromGitHub {
     owner = "NixOS";
     repo = "nixpkgs";
-    rev = "a0aeb23";
-    sha256 = "04dgg0f2839c1kvlhc45hcksmjzr8a22q1bgfnrx71935ilxl33d";
-  }){}
+    rev = "4dd5c93998da55002fdec1c715c680531420381c";
+    sha256 = "06paxakic36nbdnwkkb1094fzp3lpzxxb1r57gmb3py6pb6xrcnh";
+  }
 , haddock ? true
 }:
 let
-  inherit (pkgs.haskell.lib) buildFromSdist enableCabalFlag sdistTarball buildStrictly;
-  inherit (pkgs.haskell.packages) ghc802;
-  ghcjs = pkgs.haskell.packages.ghcjsHEAD.override {
-     overrides = self: super: {
-       jsaddle-warp = super.callPackage ./jsaddle-warp-ghcjs.nix {};
-     };
+  overrides = pkgs: {
+    haskell = pkgs.haskell // {
+      packages = pkgs.haskell.packages // {
+        ghc864 = pkgs.haskell.packages.ghc864.override {
+          overrides = self: super: with pkgs.haskell.lib; {
+            happy = dontCheck (super.callHackage "happy" "1.19.9" {});
+            mkDerivation = args: super.mkDerivation (args // {
+              enableLibraryProfiling = false;
+              doCheck = false;
+              doHaddock = false;
+            });
+          };
+        };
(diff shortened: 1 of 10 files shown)


dmjio:
cc @cocreature @FPtje @tysonzero @parthshah31, considering the following:

- [ ] Bumping nixpkgs pin to `ghcjs8.6`
- [ ] Refactoring `jsbits/*.js` to allow advanced optimizations
- [ ] Updating tests to reflect advanced optimizations changes.

Would appreciate a review, the `yarn` tests pass, and `todo-mvc` seems to pass a sanity test.

dmjio:
```
------------------|----------|----------|----------|----------|-------------------|
File              |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------------|----------|----------|----------|----------|-------------------|
All files         |    99.53 |    97.62 |      100 |    99.41 |                   |
 diff.cc.js       |    99.45 |    99.07 |      100 |     99.4 |               158 |
 isomorphic.cc.js |      100 |    88.89 |      100 |      100 |                 1 |
------------------|----------|----------|----------|----------|-------------------|
Test Suites: 1 passed, 1 total
Tests:       49 passed, 49 total
Snapshots:   0 total
Time:        4.635s
Ran all test suites.
Done in 5.22s.
```

default.nix
```diff
@@ -1,32 +1,53 @@
...
     doHaddock = haddock;
     postInstall = ''
       mkdir -p $out/bin/mario.jsexe/imgs
       cp -r ${drv.src}/examples/mario/imgs $out/bin/mario.jsexe/
       cp ${drv.src}/examples/xhr/index.html $out/bin/xhr.jsexe/
-      ${closurecompiler}/bin/closure-compiler $out/bin/todo-mvc.jsexe/all.js > $out/bin/todo-mvc.jsexe/min.js
-      rm $out/bin/todo-mvc.jsexe/all.js
-      mv $out/bin/todo-mvc.jsexe/min.js $out/bin/todo-mvc.jsexe/all.js
+      ${closurecompiler}/bin/closure-compiler --compilation_level ADVANCED_OPTIMIZATIONS \
+        --jscomp_off=checkVars \
```
FPtje: You might want to document this in e.g. the README. Which closure-compiler options should you run with and why?

tests/Main.hs
```diff
@@ -95,7 +95,7 @@ propTest prop = do
     GaveUp { output = o } -> (False, o)
     Failure { output = o } -> (False, o)
     NoExpectedFailure { output = o } -> (False, o)
-    InsufficientCoverage { output = o } -> (False, o)
+--  InsufficientCoverage { output = o } -> (False, o)
```
FPtje: Is commenting out the right thing, or should this code be removed?

default.nix
dmjio: yea, good call.

tests/Main.hs
dmjio: Probably removing altogether.

dmjio:
@FPtje I think the issue with that error above is that the `foreign import`'s need to now be cognizant of the fact they're accessing the `window` object. I've pushed a commit to reconcile this fact, care to try with this hash? `757c31a`

default.nix
dmjio: But the GHCJS docs do this too, so maybe simply a link to this reference: https://github.com/ghcjs/ghcjs/wiki/Deployment

FPtje:
Latest version confirmed working with isomorphic and jswidgets example, with modifications: I needed to string index all of my own FFI functions.

FPtje:
Also confirmed working when not using (advanced) optimisations in both aforementioned examples.

dmjio:
Examples pushed (`s3.sh`), confirmed working, and all tests pass for `diff.js` for both `ADVANCED` and omitted optimization.

PR #494, opened 2019-07-10
fix missing import

juliendehos:
Fix the following error when compiling miso:
```
$ nix-build
...
Building source dist for miso-0.21.2.0...
Preprocessing library miso-0.21.2.0...
Source tarball created: dist/miso-0.21.2.0.tar.gz
running tests
haddockPhase
installing
post-installation fixup
[ 8 of 21] Compiling Miso.Html.Event  ( src/Miso/Html/Event.hs, dist/build/Miso/Html/Event.o )
[ 9 of 21] Compiling Miso.Html.Property ( src/Miso/Html/Property.hs, dist/build/Miso/Html/Property.o )

src/Miso/Html/Property.hs:391:31: error:
    • Variable not in scope: (<>) :: [Char] -> MisoString -> MisoString
    • Perhaps you meant one of these:
        ‘<$’ (imported from Prelude), ‘<$>’ (imported from Prelude),
        ‘*>’ (imported from Prelude)
      Perhaps you want to add ‘<>’ to the import list in the import of
      ‘Miso.String’ (src/Miso/Html/Property.hs:122:1-54).
```

### DIFF
--- a/src/Miso/Html/Property.hs
+++ b/src/Miso/Html/Property.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE CPP #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Miso.Html.Property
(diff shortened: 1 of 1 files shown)


dmjio:
`base` has recently included `(<>)` in its list of exports, which would cause a collision with this import. Would it be possible to account for this by adding a CPP pragma that does not import `Data.Monoid` if the version of base is greater than the version that began to re-export `(<>)` ?

dmjio:
Seems to work great. Thank you so much! @juliendehos .

PR #495, opened 2019-07-12
remove TypeApplications for building with GHC 7.10

juliendehos:
Hi, Not sure if this is very important but this PR should make Miso usable with GHC 7.10 (tested with this project https://gitlab.com/juliendehos/miso-test).

### DIFF
--- a/ghc-src/Miso/String.hs
+++ b/ghc-src/Miso/String.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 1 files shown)


dmjio:
@juliendehos, definitely important. Thank you !

Also, that repo seems like it would be good for a continuous integration setup (to test different ghc versions against miso). We could possibly cache all the deps. (from a few nixpkgs repos) into an s3 bucket. The CI builds could fetch the deps. from the cache.

juliendehos:
Indeed, it would be great to have a few pinned configs + cache. I sometimes use a pre-built docker image (e.g, https://gitlab.com/juliendehos/miso-xhr/tree/master/dockerhub) but it's quite heavy to run. In the `miso-test` repo, the example is very simple so it may be useful to include other deps too.

PR #513, opened 2019-07-20
Add support for cross compiling to iOS.

dmjio:
Refactors jsaddle code, adds ARM cross compiler.


dmjio:
All that's left should be replacing `file-embed` TH with data sources config equivalent. Pushed ARM deps to cachix.

PR #531, opened 2019-10-06
Add commercial users section

dmjio:
cc @basvandijk @tysonzero @FPtje @parthshah31


FPtje:
LumiGuide without space 😋

PR #538, opened 2019-10-16
Fix live reload entr command example in readme

razvan-flavius-panda:
The command for live reload example is wrong:

    ag -l | entr 'cabal build'

In above command it thinks that `cabal build` is an executable so it fails with error:

    entr: exec cabal build: No such file or directory


dmjio:
Thanks :+1:

PR #540, opened 2019-10-20
Fix sample-app-jsaddle for users without a global installation of cabal

srid:


dmjio:
Solid, thanks @srid  👍

PR #544, opened 2019-11-05
Fix the transition application

y-taka-23:
Results in compilation errors. Surely because they are just copied from the Sample application above.


dmjio:
Thanks @y-taka-23 :beers:

PR #548, opened 2019-11-16
Adjust keys patch to work correctly when sorting large keyed lists.

dmjio:
cc @y-taka-23


dmjio:
@y-taka-23 would it be possible for you to join the slack? Would like to show you some videos of the keys patch.

PR #552, opened 2019-12-04
Doc fixes

jhrcek:
I'd very much like to submit some more improvements to the documentation, but I failed to build the haddocks locally. Can you share a command you're using to build the documentation locally?

### DIFF
--- a/examples/svg/Main.hs
+++ b/examples/svg/Main.hs
@@ -55,8 +55,8 @@ viewModel (Model (x,y)) =
   div_ [ ] [
     svg_ [ style_ $ M.fromList [ ("border-style", "solid")
                                , ("height", "700px")
+                               , ("width", "100%")
                                ]
-         , width_ "auto"
          , onTouchMove HandleTouch
        ] [
      g_ [] [
--- a/frontend-src/Miso/Types.hs
+++ b/frontend-src/Miso/Types.hs
@@ -40,18 +40,19 @@ data App model action = App
   { model :: model
   -- ^ initial model
   , update :: action -> model -> Effect action model
-  -- ^ Function to update model, optionally provide effects.
+  -- ^ Function to update model, optionally providing effects.
   --   See the 'Transition' monad for succinctly expressing model transitions.
   , view :: model -> View action
   -- ^ Function to draw `View`
   , subs :: [ Sub action ]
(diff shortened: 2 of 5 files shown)


examples/svg/Main.hs
```diff
@@ -55,8 +55,8 @@ viewModel (Model (x,y)) =
   div_ [ ] [
     svg_ [ style_ $ M.fromList [ ("border-style", "solid")
                                , ("height", "700px")
+                               , ("width", "100%")
                                ]
-         , width_ "auto"
```
jhrcek: Not sure what was the intention behind `width "auto"` but in google chrome it renders across entire screen width, and throws this error in js console:
```bash
svg example : lib.js:1811 Error: <svg> attribute width: Expected length, "auto".
```
Whereas in firefox it shows only narrow, but tall svg element and the view looks broken.

frontend-src/Miso/Types.hs
```diff
@@ -40,18 +40,19 @@ data App model action = App
...
+  -- ^ Function to update model, optionally providing effects.
   --   See the 'Transition' monad for succinctly expressing model transitions.
   , view :: model -> View action
   -- ^ Function to draw `View`
   , subs :: [ Sub action ]
   -- ^ List of subscriptions to run during application lifetime
   , events :: M.Map MisoString Bool
-  -- ^ List of delegated events that the body element will listen for
+  -- ^ List of delegated events that the body element will listen for.
+  --   You can start with 'Miso.Event.Types.defaultEvents' and modify as needed.
```
jhrcek: Can you please clarify what's is the difference in behavior between some event name 
- not being in the map at all
- being in the map with True value
- being in the map with False value?

I think this kind of information should be added to docs as well.

dmjio: - Not being in the map means it is not being listened for at all and will therefore never fire.
- In the map with `True` means the even will fire and will use `captureMode`
- In the map with `False` means the even will fire and will *not* use `captureMode`

dmjio:
@jhrcek 👋 Hi! I would absolutely /love/ for better documentation, and would be happy to share knowledge of how `miso` works in this regard.

Huge thanks for the contributions. To follow up would you like to join the slack / IRC ?

PR #553, opened 2019-12-07
Improve haddock coverage of things exported from Miso module

jhrcek:
@dmjio is this the kind of docs you're looking for? Maybe this kind of trivial docs are not really useful for people forged by the fire of web platform, but for someone like me (relatively inexperienced in web development) it would be useful :)

Also there are bunch of functions exported from Miso module, about which I have no idea as to their purpose, like callbackToJSVal, objectToJSVal, ghcjsPure, swapCallbacks, releaseCallbacks, registerCallback. I'll dive into the source code and hopefully understand their usefulness. I'd like to add some short code examples to illustrate the usage of some non-obvious functions.

### DIFF
--- a/ghcjs-ffi/Miso/FFI.hs
+++ b/ghcjs-ffi/Miso/FFI.hs
@@ -76,6 +76,7 @@ import qualified JavaScript.Object.Internal as OI
 -- resolves to the `JSM` type defined in `jsaddle`.
 type JSM = IO
 
+-- | Run given `JSM` action asynchronously, in a separate thread.
 forkJSM :: JSM () -> JSM ()
 forkJSM a = () <$ forkIO a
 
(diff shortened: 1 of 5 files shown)


ghcjs-ffi/Miso/FFI.hs
```diff
@@ -88,21 +89,28 @@ objectToJSVal = pure . jsval
 ghcjsPure :: a -> JSM a
 ghcjsPure = pure
 
+-- | Forces execution of pending asyncronous code
```
jhrcek: Not sure if this is right (just copied if over from jsaddle docs). Does it mean that this only have an effect when running with jsaddle?

jhrcek:
Couple more things I'm not certain about:
The doc comments in FFI files are duplicated (I copy pasted them), not sure if that's desirable.

Locally I was building and checking the docs using
`nix-shell -A miso-ghcjs.env --run 'cabal v1-haddock'`
Which I assume gives the same result as what's published on https://haddocks.haskell-miso.org/index.html

But I was not able to figure out a command to build the docks as they're [on hackage](http://hackage.haskell.org/package/miso-1.4.0.0). Those have some modules greyed-out. I guess those are "frontend-only" modules, that are not available for ghc. Is there a command to build the docs as they're on hackage, or are the greyed-out modules a result of uploading to hackage?

PR #563, opened 2020-02-10
jsaddle example: better reload script

turboMaCk:
I think this is a better way to provide `reload` script in nix-shell environment. Even though technically speaking I see nothing bad about the old way I think this one is more idiomatic nix way of accomplishing it. Advantage is that with this change the script is created (and cached) in nix store though for something as simple as this it doesn't matter that much.

### DIFF
--- a/sample-app-jsaddle/shell.nix
+++ b/sample-app-jsaddle/shell.nix
@@ -1,9 +1,10 @@
 with (import ./default.nix);
-dev.env.overrideAttrs (old: {
-  shellHook = ''
-    function reload () {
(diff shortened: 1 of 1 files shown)


dmjio:
The advantage of the original is that it is usable as a bash function since it is defined in a `shellHook`. Otherwise, the code looks identical.

`nix-shell --run repl`

turboMaCk:
@dmjio yes the above will create script in nix store which is in PATH.

```
[nix-shell] which reload
/nix/store/zhm3jkhs3azcqaihq3xgip8cbiy0qzgc-reload/bin/reload
```

In my opinion it also makes it clearer how one can go about adding anything other than shell scripts without wrapping it to shell function.

Issue #492, opened 2019-07-05
Split ToMisoString class?


noinia:
The ToMisoString class currently has two functions: 
`toMisoString :: str -> MisoString`
and 
`fromMisoString :: MisoString -> str `

and there are instances for non-string types such as Int, Double, and Float. This seems wrong to me, as there are clearly MisoStrings that cannot be converted properly into Ints, Doubles, or Floats. It is clearly nice to be able to easily serialize such types into MisoStrings when building a view [1]. So I think it would be much cleaner to have separate 'ToMisoString' and 'FromMisoString' typeclasses for the toMisoString and fromMisoString functions, respectively.

[1] this is actually how I ran into this issue; serializing things such as RGB value from the Colour package into a misoString.

dmjio:
`fromMisoString` was meant to be the analog of `read :: Read a => String -> a` for `String`.

> there are instances for non-string types such as Int, Double, and Float. This seems wrong to me, as there are clearly MisoStrings that cannot be converted properly into Ints, Doubles, or Floats.

the same logic would apply to `String` as well, `Read` class simple throws an exception. We could do `fromMisoString :: MisoString -> Either String str` (similar to `Text.Read`) and be verbose about why the conversion does not exist. I think the former was chosen out of expediency.

> So I think it would be much cleaner to have separate 'ToMisoString' and 'FromMisoString' typeclasses for the toMisoString and fromMisoString functions, respectively.

Doing this would definitely remove the need to prove the law that 
`\x -> id x === fromMisoString (toMisoString x)`, `aeson` takes a similar approach with separating out typeclasses for these things. It's definitely worth considering. I'll consider it. If you make a PR I'll be much more apt to accept.

dmjio:
solved in https://github.com/dmjio/miso/commit/7881ce9364beb00b017996b6f9e777284e6adcd5

PR #564, opened 2020-02-13
split off a FromMisoString class from the ToMisoString class ( adresses #492)

noinia:
WIP: don't merge yet

This partly addresses #492. Essentially, I created a new 'FromMisoString' class that has a 'fromMisoStringEither' function that tries to read a MisoString into an Either String a. Still todo: 

1. for ghcjs I added a 'jsStringToDoubleEither' function whose function it is to try to parse double. The current implementation is kind of silly, as it just directly uses 'jsStringToDoubleEither' and wraps it in a 'Right'. Clearly, that is not yet the right thing to do. (But I will have to figure out how to do the right thing there). 

2. Trying to parse a, say Int, from a nonsensical string the error returned in the Left is now simply "Prelude.read: no parse". Maybe we want a better (= more specific) message there.

3. Update docs where necessary.

Before going through those steps (in particular 1 will require some figuring out for me), let me ask: Do you think this is a reasonable solution?

### DIFF
--- a/frontend-src/Miso/String.hs
+++ b/frontend-src/Miso/String.hs
@@ -116,5 +116,6 @@ instance FromMisoString Word where
   fromMisoStringEither = fmap round . jsStringToDoubleEither
 
 jsStringToDoubleEither :: JSString -> Either String Double
-jsStringToDoubleEither = Right . jsStringToDouble
--- TODO
+jsStringToDoubleEither s = let d = jsStringToDouble s
+                           in if isNaN d then Left "jsStringToDoubleEither: parse failed"
+                                         else Right d
--- a/frontend-src/Miso/String.hs
+++ b/frontend-src/Miso/String.hs
@@ -26,6 +26,8 @@ module Miso.String (
 
 #ifndef JSADDLE
 import           Data.Aeson
+import           Data.Char
+
 #endif
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Lazy as BL
@@ -36,13 +38,15 @@ import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.Lazy as LT
 import qualified Data.Text.Lazy.Encoding as LT
-
+import           Prelude hiding (foldr)
 import           Miso.FFI
 
 -- | String type swappable based on compiler
 type MisoString = JSString
 
 #ifndef JSADDLE
+
+
 -- | `ToJSON` for `MisoString`
 instance ToJSON MisoString where
   toJSON = String . textFromJSString
@@ -111,11 +115,25 @@ instance FromMisoString Float where
 instance FromMisoString Double where
   fromMisoStringEither = jsStringToDoubleEither
 instance FromMisoString Int where
-  fromMisoStringEither = fmap round . jsStringToDoubleEither
+  fromMisoStringEither = parseInt
 instance FromMisoString Word where
-  fromMisoStringEither = fmap round . jsStringToDoubleEither
+  fromMisoStringEither = parseWord
 
 jsStringToDoubleEither :: JSString -> Either String Double
 jsStringToDoubleEither s = let d = jsStringToDouble s
                            in if isNaN d then Left "jsStringToDoubleEither: parse failed"
                                          else Right d
+
+
+parseWord :: MisoString -> Either String Word
+parseWord = fmap snd . foldr k (Right (1,0))
+  where
+    k c acc = do (l,x) <- acc
+                 y     <- if isDigit c then Right (fromIntegral $ digitToInt c)
+                                       else Left "parseWord: parse error"
+                 pure $ (l*10,l*y + x)
+
+parseInt   :: MisoString -> Either String Int
+parseInt s = case uncons s of
+               Just ('-',s') -> ((-1)*) . fromIntegral <$> parseWord s'
+               _             ->           fromIntegral <$> parseWord s
--- a/frontend-src/Miso/String.hs
+++ b/frontend-src/Miso/String.hs
@@ -16,6 +16,8 @@
 ----------------------------------------------------------------------------
 module Miso.String (
     ToMisoString (..)
+  , FromMisoString (..)
+  , fromMisoString
   , MisoString
   , module Data.JSString
   , module Data.Monoid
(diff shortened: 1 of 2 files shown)


dmjio:
Sounds reasonable to me, definitely warrants a major version bump, but should probably happen. `fromMisoString` was added to `ToMisoString` a little haphazardly. Am for these changs.

dmjio:
@noinia do you still want to move forward with this change?

noinia:
Yep that would be nice. I just made it so that it actually tries to parse doubles and gives an Left <some error message> if that goes wrong. I also checked the docs if something is referring 'fromMisoString', but that does not seem to be the case. So I think it should be ok now :).

noinia:
By the way. I kept the semantics of parsing Ints as it was. But admittedly that is a bit strange; i.e. on JS it parses '5.7' successfully into '6'. Which I would call wrong. I didn't really want to change those semantics though.

dmjio:
hmm, yea that is a bit strange, maybe we can use 
```
foreign import javascript unsafe "$r = parseInt($1)"  parseInt :: MisoString -> Int
```
This will treat `5.7` as `5`.

noinia:
According to https://www.w3schools.com/jsref/jsref_parseint.asp parseInt will still parse successfully if the number is just a prefix of the string, so "50foobar" will parse to 50, which still seems wrong to me. Something like

```
jsStringToIntegralEither   :: Integral b => String -> Either String b
jsStringToIntegralEither s = do d <- jsStringToDoubleEither s
                                let (i,r) = properFraction d
                                if r == 0 then Right i
                                          else Left "parse error"
```

should do, maybe there is a cleaner way of writing that though.

noinia:
Although apparently that still does something weird for Word. So there we would have to check if d > 0 before taking the properFraction I guess.

noinia:
Actually, all of that is very convoluted. I think I would just write: 

```haskell
parseWord :: String -> Either String Word
parseWord = fmap snd . foldr k (Right (1,0))
  where
    k c acc = do (l,x) <- acc
                 y     <- if isDigit c then Right (fromIntegral $ digitToInt c)
                                       else Left "parse error"
                 pure $ (l+1,l*y + x)

parseInt         :: String -> Either String Int 
parseInt ('-':s) = ((-1)*) . fromIntegral <$> parseWord s
parseInt s       = fromIntegral <$> parseWord s
```

dmjio:
My only pause for concern is that it might be much slower to do this in Haskell rather than JS. In fact @FPtje might be able to speak to the performance of this code, since I believe he found a performance bottleneck related to `read` and `show`. Can you opine on this @FPtje ?

FPtje:
Iirc I indeed found a serious bottleneck in read and show. I think it was `show` particularly where I got problems. Thin of an SVG with thousands of figures, each with a dozen or so numbers that needed to be converted to strings. The optimisation of using specialised functions made our application an order of magnitude faster. Think of going from about 30 seconds to sub second rendering times.

I'm not sure whether we use the `fromMisoString` function though, I would have to look.

FPtje:
We do use it, but mostly in Actions. That might have to perform well if those actions are fired often, but I reckon it's less of a problem than conversions in the view.

noinia:
@dmjio are you also worried about the performance of the version using fromJSDoubleEither and properFraction above? If so, is there some way of benchmarking that reliably? 

By the way 1: I realized that the last implementation of parseWord I gave actually also still has a bug as it incorrectly parses "" into Right 0. The version below fixes that, and should furthermore be slightly faster as it avoids allocating tuples. 

```haskell
parseWord   :: MisoString -> Either String Word
parseWord s = case uncons s of
                Nothing     -> Left "parseWord: parse error"
                Just (c,s') -> foldl' k (pDigit c) s'
  where
    pDigit c | isDigit c = Right . fromIntegral . digitToInt $ c
             | otherwise = Left "parseWord: parse error"
    k ea c = (\a x -> 10*a + x) <$> ea <*> pDigit c
```

By the way 2: I think the parseWord above and the version using fromJSDoubleEither still have slightly different semantics, as the one going through doubles allows you to parse "1e3" as 1000, whereas the explicit parser above returns a parse error. Is there a preference for either version?

dmjio:
@noinia I'm going to add some benchmarks for this. Regenerating the view should be as cheap as possible, since it happens it a lot.

dmjio:
@noinia benchmarks have shown that your changes are in every case significantly faster than the existing implementation (anywhere from 10x to 25x faster). 

The test bed consisted of parsing a single `MisoString` into an `Int` using the current and proposed implementations. It was timed with a before-and-after comparison using `Performance.now()`. To get more test samples I just refresh the browser. 

Your changes parse consistently in less than `0.005ms`. All other implementations are at least `0.05 ms`. Code below.

```haskell
{-# LANGUAGE BangPatterns      #-}                                                                                                                                                       
                                                                                                                                                                                         
module Main where                                                                                                                                                                        
                                                                                                                                                                                         
import Miso                                                                                                                                                                              
import Miso.String                                                                                                                                                                       
                                                                                                                                                                                         
import Data.Char                                                                                                                                                                         
                                                                                                                                                                                         
foreign import javascript unsafe "$r = Number($1);"                                                                                                                                      
  oldJsStringToDouble :: JSString -> Double                                                                                                                                              
                                                                                                                                                                                         
toInt :: MisoString -> Int                                                                                                                                                               
toInt = round . oldJsStringToDouble                                                                                                                                                      
                                                                                                                                                                                         
parseInt   :: MisoString -> Either String Int                                                                                                                                            
parseInt s = case uncons s of                                                                                                                                                            
               Just ('-',s') -> ((-1)*) . fromIntegral <$> parseWord s'                                                                                                                  
               _             ->           fromIntegral <$> parseWord s                                                                                                                   
                                                                                                                                                                                         
parseWord   :: MisoString -> Either String Word                                                                                                                                          
parseWord s = case uncons s of                                                                                                                                                           
                Nothing     -> Left "parseWord: parse error"                                                                                                                             
                Just (c,s') -> foldl' k (pDigit c) s'                                                                                                                                    
  where                                                                                                                                                                                  
    pDigit c | isDigit c = Right . fromIntegral . digitToInt $ c                                                                                                                         
             | otherwise = Left "parseWord: parse error"                                                                                                                                 
    k ea c = (\a x -> 10*a + x) <$> ea <*> pDigit c                                                                                                                                      
                                                                                                                                                                                         
main :: IO ()                                                                                                                                                                            
main = new >> old                                                                                                                                                                        
                                                                                                                                                                                         
example :: MisoString                                                                                                                                                                    
example = "2147483647"                                                                                                                                                                   
                                                                                                                                                                                         
old, new :: IO ()                                                                                                                                                                        
old = do                                                                                                                                                                                 
  consoleLog ("Old" :: MisoString)                                                                                                                                                       
  start <- now                                                                                                                                                                           
  let !a = toInt example                                                                                                                                                                 
  end <- now                                                                                                                                                                             
  consoleLog $ "Elapsed: " <> ms (end - start)                                                                                                                                           
  consoleLog (ms (a :: Int))                                                                                                                                                             
                                                                                                                                                                                         
new = do                                                                                                                                                                                 
  consoleLog ("New" :: MisoString)                                                                                                                                                       
  start <- now                                                                                                                                                                           
  let Right !a = parseInt example                                                                                                                                                        
  end <- now                                                                                                                                                                             
  consoleLog $ "Elapsed: " <> ms (end - start)                                                                                                                                           
  consoleLog (ms a) 
```

<img width="258" alt="Screen Shot 2020-03-24 at 8 49 22 PM" src="https://user-images.githubusercontent.com/875324/77490644-d8432580-6e11-11ea-9ed5-bd1719d64a6e.png">

I added `parseInt` as well, and it parses stably around `0.04ms - 0.07ms`.

Awesome work. This is a great addition.

noinia:
Cool, thanks!

PR #572, opened 2020-04-04
Fix error in code in README.md

janeymunoz:
Command as written in README causes erroneous behavior:
```
[nix-shell:~/]$ nix-shell -A env --run 'cabal configure --ghcjs && cabal build`
>
```

After edit to command, desired behavior occurs:
```
[nix-shell:~/]$ nix-shell -A env --run 'cabal configure --ghcjs && cabal build'
'cabal.project.local' file already exists. Now overwriting it.
Build profile: -w ghcjs-8.6.0.1 -O1
In order, the following would be built (use -v for more details):
 ...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ nix-shell -A env --run 'ghcjs-pkg list'
 
 To build the project with `cabal` after entering the `nix-shell`
 ```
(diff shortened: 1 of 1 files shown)


dmjio:
Nice catch, thanks.

PR #583, opened 2020-05-19
Support some pass-through nixpkgs arguments.

dhess:
Among other things, this makes it possible to build Miso for multiple
platforms with Hydra.


dmjio:
@dhess Thanks !

PR #586, opened 2020-06-03
Make no element tags when appropriate.

dmjio:
cc @niteria


niteria:
Wow, that's quick. Thank you!

Issue #589, opened 2020-06-04
Empty text nodes seem to cause problems in isomorphic Miso


niteria:
**Describe the bug**
I have an example where changing `""` to `" "` makes a difference between the app working and a crash in a browser.

Here's a crashing version: https://github.com/niteria/miso-isomorphic-example/commit/a22975c537e9edd9ca4b236d3ee132c4340f3c06
Here's a patch that makes it work: https://github.com/niteria/miso-isomorphic-example/commit/22ebdaa47d0830a0131afa1130557f273e60d656

**To Reproduce**
```
git clone https://github.com/niteria/miso-isomorphic-example && cd miso-isomorphic-example
git checkout empty-text-node-problem
(cd $(nix-build) && bin/server)
# look at localhost:3003 in the browser
# click the "+" button
# the app stops responding after that and there's an error in the browser console
```

Browser console output:
```
uncaught exception in Haskell main thread: TypeError: b.domRef is undefined all.js:215:83
window.diffTextNodes@http://localhost:3003/static/all.js:680:342
window.diff@http://localhost:3003/static/all.js:679:180
window.diffChildren@http://localhost:3003/static/all.js:684:400
window.populate@http://localhost:3003/static/all.js:682:174
window.diffVNodes@http://localhost:3003/static/all.js:682:313
window.diff@http://localhost:3003/static/all.js:679:223
window.diffChildren@http://localhost:3003/static/all.js:684:400
window.populate@http://localhost:3003/static/all.js:682:174
window.diffVNodes@http://localhost:3003/static/all.js:682:313
window.diff@http://localhost:3003/static/all.js:679:223
h$$aaW@http://localhost:3003/static/all.js:1878:96
h$runThreadSlice@http://localhost:3003/static/all.js:255:213
h$runThreadSliceCatch@http://localhost:3003/static/all.js:254:46
h$mainLoop@http://localhost:3003/static/all.js:253:163
h$mkForeignCallback/<@http://localhost:3003/static/all.js:275:238
all.js:215:83
uncaught exception: Thread aborted, exit code: 0 all.js:268:401
```

**Expected behavior**
No crash.

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version: 76.0.1

**Additional context**
This is a minimized example from a problem I run into in a real app. It's not doing anything interesting.

niteria:
[...]

Issue #520, opened 2019-08-14
console.warn on consecutive text nodes


dmjio:
When detecting sibling text nodes, `console.warn` the DOM and VDOM may not be in synchrony.

dmjio:
https://github.com/dmjio/miso/pull/591

Issue #588, opened 2020-06-03
Diagnostic info when Miso.miso fails


niteria:
**Is your feature request related to a problem? Please describe.**
This is based on my experience diagnosing #585 and #587.

When a framework user makes an error and the view renders inconsistently between the server and the client in an isomorphic app the only feedback that the user gets is some errors in the console.

From my short experience, it's quite easy to introduce such a bug, so it'd be nice to have more diagnostic info. 

**Describe the solution you'd like**
I think there could be a flag to `Miso.miso` that makes it perform a sanity check on start-up, basically rerendering the view with the model and checking if that's what was shipped from the server. Perhaps the diffing algorithm can be used to print the difference to the console.

**Describe alternatives you've considered**
I only had a brief look at the diffing code, so I might be wrong about this. From what I understand the diffing algorithm assumes that all the DOM nodes rendered with the old model exist. Perhaps it can print some diagnostics when that assumption is broken.

I'm a bit more skeptical of this because it'd an ongoing cost of extra checks instead of a one-time check at startup with the solution proposed in the previous section.

dmjio:
@niteria it should fallback to rendering everything if it determines the view can't render, and `warn` about doing so.

https://github.com/dmjio/miso/blob/master/jsbits/isomorphic.js#L6

Which version of miso are you using?

niteria:
I'm a bit confused then.
The reproducer for #585 used miso `1.5.2` which has an exact same definition for that function.

If I understand correctly the only problem there was that the server sent HTML that was interpreted as two `<br>` tags instead of one. But the console didn't print `"Could not copy DOM into virtual DOM, falling back to diff"` and the app crashed in the `walk` function.

I see that I incorrectly assumed that the `walk` function needs the DOM to agree with the virtual tree created from the Model on the JS side.

If I discard that assumption then going back to the crash in #585 `walk` crashed in:
https://github.com/dmjio/miso/blob/master/jsbits/isomorphic.js#L32, `domChild.nodeType` (sorry the stack trace from #585 had variable names that went through the minimizer).

So at that point in time for the node that `walk` was processing there were fewer DOM nodes than the VDOM nodes.

That makes sense, we were processing this DOM:
```
<body data-gr-c-s-loaded="true"><div><div><button>-</button>0<br><br><button>+</button></div><button>Go to /flipped</button></div></body>
```

And the VDOM was:
```
[...]

Issue #587, opened 2020-06-03
Route Network.URIs inconsistent between server and client in isomorphic example


niteria:
**Describe the bug**
This is similar to https://github.com/dmjio/miso/issues/585.
If I modify the view code to use `Show Network.URI` instance the application hangs on start up.
Change for reference: https://github.com/niteria/miso-isomorphic-example/commit/3e9eaa41c9943db01d8d088d8a00391181d80c1d

**To Reproduce**
```
git clone https://github.com/niteria/miso-isomorphic-example && cd miso-isomorphic-example
git checkout network-uri-inconsistent
(cd $(nix-build) && bin/server)
# look at localhost:3003/flipped or localhost:3003/ in the browser, doesn't work anymore. /flipped better illustrates the problem
```

Message from the browser console:
```
Could not copy DOM into virtual DOM, falling back to diff
```

**Expected behavior**
Normal behavior from the example app, with extra text somewhere.

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version: 76.0.1

**Additional context**
I modified the app to use `Miso.startApp` instead of `Miso.miso` like so: https://github.com/niteria/miso-isomorphic-example/commit/d8a00d97238868dcc74bd7e178535d6b8c9d4cc3.

Running the code you can see that the browser `Show`s the flipped route as `http://localhost:3003/flipped` and the server `Show`s it as `flipped`.

dmjio:
this should be fixed since the node collapse, re-open if not.

PR #591, opened 2020-06-08
Collapse sibling text nodes.

dmjio:
@niteria @FPtje @tysonzero 

This preserves the isomorphism between the DOM and HTML by normalizing both structures by collapsing sibling text nodes. It should* solve the "consecutive text node" problem during pre-rendering.

Solves #589 #520 .

### DIFF
--- a/frontend-src/Miso/Html/Internal.hs
+++ b/frontend-src/Miso/Html/Internal.hs
@@ -131,7 +131,6 @@ node ns tag key attrs kids = View $ \sink -> do
   set "key" key vnode
   setAttrs vnode sink
   flip (set "children") vnode
-    =<< collapseSiblingTextNodes
     =<< ghcjsPure . jsval
     =<< setKids sink
   pure $ VTree vnode
--- a/frontend-src/Miso/Html/Internal.hs
+++ b/frontend-src/Miso/Html/Internal.hs
@@ -130,7 +130,10 @@ node ns tag key attrs kids = View $ \sink -> do
   set "tag" tag vnode
   set "key" key vnode
   setAttrs vnode sink
-  flip (set "children") vnode =<< ghcjsPure . jsval =<< setKids sink
+  flip (set "children") vnode
+    =<< collapseSiblingTextNodes
+    =<< ghcjsPure . jsval
+    =<< setKids sink
   pure $ VTree vnode
     where
       setAttrs vnode sink =
--- a/ghc-src/Miso/Html/Internal.hs
+++ b/ghc-src/Miso/Html/Internal.hs
@@ -124,7 +124,7 @@ instance L.ToHtml (VTree action) where
 collapseSiblingTextNodes :: [VTree a] -> [VTree a]
 collapseSiblingTextNodes [] = []
 collapseSiblingTextNodes (VText x : VText y : xs) =
-  collapseSiblingTextNodes (VText (x <> " " <> y) : xs)
+  collapseSiblingTextNodes (VText (x <> y) : xs)
 collapseSiblingTextNodes (x:xs) =
   x : collapseSiblingTextNodes xs
 
--- a/ghc-src/Miso/Html/Internal.hs
+++ b/ghc-src/Miso/Html/Internal.hs
@@ -114,7 +114,19 @@ instance L.ToHtml (VTree action) where
                      , "autocomplete"
                      ]
         toTag = T.toLower
-        kids = foldMap L.toHtml vChildren
+        kids
+          = foldMap L.toHtml
+          . V.fromList
+          . collapseSiblingTextNodes
+          . V.toList
+          $ vChildren
+
+collapseSiblingTextNodes :: [VTree a] -> [VTree a]
+collapseSiblingTextNodes [] = []
+collapseSiblingTextNodes (VText x : VText y : xs) =
+  collapseSiblingTextNodes (VText (x <> " " <> y) : xs)
+collapseSiblingTextNodes (x:xs) =
+  x : collapseSiblingTextNodes xs
 
 -- | Helper for turning JSON into Text
 -- Object, Array and Null are kind of non-sensical here
--- a/jsbits/isomorphic.js
+++ b/jsbits/isomorphic.js
@@ -14,23 +14,31 @@ window['collapseSiblingTextNodes'] = function collapseSiblingTextNodes(vs) {
 
 window['copyDOMIntoVTree'] = function copyDOMIntoVTree(debug,mountPoint, vtree, doc) {
   if (!doc) { doc = window.document; }
-  var mountChildIdx = 0;
-  // If script tags are rendered first in body, skip them.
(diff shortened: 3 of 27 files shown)


ghc-src/Miso/Html/Internal.hs
```diff
@@ -114,7 +114,19 @@ instance L.ToHtml (VTree action) where
...
+          = foldMap L.toHtml
+          . V.fromList
+          . collapseSiblingTextNodes
+          . V.toList
+          $ vChildren
+
+collapseSiblingTextNodes :: [VTree a] -> [VTree a]
+collapseSiblingTextNodes [] = []
+collapseSiblingTextNodes (VText x : VText y : xs) =
+  collapseSiblingTextNodes (VText (x <> " " <> y) : xs)
```
FPtje, line 127: Should there really be a space in between? I would expect multiple text nodes to be concatenated directly.

jsbits/util.js
```diff
@@ -12,3 +12,16 @@ window['callBlur'] = function callBlur(id) {
     if (ele && ele.blur) ele.blur()
   }, 50);
 }
+
+window['collapseSiblingTextNodes'] = function collapseSiblingTextNodes(vs) {
+  if (!vs) { return []; }
+  var ax = 0, adjusted = vs.length > 0 ? [vs[0]] : [];
+  for (var ix = 1; ix < vs.length; ix++) {
+    if (adjusted[ax]['type'] === 'vtext' && vs[ix]['type'] === 'vtext') {
+	adjusted[ax]['text'] += ' ' + vs[ix]['text'];
```
FPtje, line 21: Same here

frontend-src/Miso/Html/Internal.hs
```diff
@@ -130,7 +130,10 @@ node ns tag key attrs kids = View $ \sink -> do
   set "tag" tag vnode
   set "key" key vnode
   setAttrs vnode sink
-  flip (set "children") vnode =<< ghcjsPure . jsval =<< setKids sink
+  flip (set "children") vnode
+    =<< collapseSiblingTextNodes
```
FPtje, line 134: It looks like the children are being collapsed in ghcjs before every diff. Is this necessary? Can we get away with doing it only on the initial sync up between the server sent html and first render?

I'm not sure there were diffing problems between double text nodes when not using isomorphism, so feel free to shoot this idea down immediately 😅

dmjio: This is a good idea, we shouldn't call collapse on each diff iteration.

jsbits/isomorphic.js
```diff
@@ -1,19 +1,40 @@
...
   }
+  if (debug) {
+    console.info ('Successfully prendered page');
+  }
   return true;
 }
 
-window['walk'] = function walk(vtree, node, doc) {
+window['diagnoseError'] = function diagnoseError(debug, vtree, node) {
+    if (debug) console.warn('VTree differed from node', vtree, node);
```
FPtje, line 34: I'm a bit skeptic about the `debug` variable here. The name of the variable (`debug`) does not immediately make clear *what* it debugs. Documentation has to be read.

Besides, what it does is show these debug prints. Is that something we actually expect end users to use? Is it easy enough to mismatch and hard enough to debug such mismatches? Biggest cause for mismatches AFAIK are those adjacent text nodes. That's exactly what you fix.

Perhaps we should look at a different means of debugging such mismatches, maybe some wrapper around the `miso` function?

niteria: I've used `miso` for a couple of weeks and have run into multiple different causes for the mismatches, including:

* The problem you mention - adjacent text nodes
* Different Route URIs: #587
* Bug in the framework that renders `br_` differently: #585
* My own mistakes where I didn't ensure that the Model on the server is the same that the model I use in the client

Each time I had to resort to commenting out pieces of the code to narrow it down and before the fallback code was fixed in this pull request the app would just hang with no feedback.

I believe this functionality will remain useful, I have no strong opinions on the way to provide it.

niteria: If we disagree, can we move the discussion to #588 and proceed with the main contents of this pull request? This is a really nice fix and I wouldn't want to stall it on a tangential issue.

niteria: Alternatively, we could rename `debug` to `logLevel` and have multiple logging levels: `0 - off`, `1 - error`, `2 - warn`, `3 - info`, `4 - debug`, `5 - trace`.

dmjio:
@niteria @FPtje I've added some additional handling of `mountPoint`, that will ensure `miso` isn't mounted on a text node, and that miso will play nice alongside `<script>` that exist as siblings in `<body>`. I can rename `debug` to something more specific, i.e. `logLevel`. We should in theory be ready to go now.

FPtje:
logLevel sounds good

dmjio:
@niteria @ParthShah31 @tysonzero @BenjaminBaker @FPtje

I think this is ready to go (build is red due to `nix` issues), so disregard.

This has incorporated all of the aforementioned and is working well (tests pass and integration tested on some extremely large apps cc @ParthShah31). This really makes the isomorphic functionality first class, excited for things to become more user-friendly. Any last comments, concerns, cries of outrage before we merge?

FPtje:
> cries of outrage

Well actually...

Just kidding 😂
LGTM! Nice work!

niteria:
LGTM, but I expected `logLevel` to be a number, so that if someone only wants to see the errors they'd set it to `1` and the places that print errors would do `if (logLevel >= 1) ...` (in real code `1` hidden behind a constant, but shown here for demonstration). 

If someone wants to see the the debug messages they'd set it to `4` and the places with debug messages would do `if (logLevel >= 4) ...`. The nice thing is that the debug messages also enable error messages.

Might be overkill for now since we only have debug messages, but that's kind of what I had in mind.

dmjio:
@niteria I agree with that approach, and I think we should adopt it as we add more logging facilities. For now, besides logging during prerendering, there isn't much else to log besides diffing. I didn't want to include extra JS log messages in diff.js because it would slow it down. Another use case for logging would be component mounting unmounting, but components right now are in Haskell land only, and have no js representation. So that's not a good use case for logging yet. But I agree, we can refactor add more when the need arises.

tysonzero:
Would it be possible to have a separate log level that only shows warnings and not successes.

We won't really want to log "Successfully prendered page", particularly in production, but keeping the warnings around when things go wrong is definitely useful.

I know persistent uses [LogLevel](https://hackage.haskell.org/package/monad-logger-0.3.32/docs/Control-Monad-Logger.html#t:LogLevel) for this kind of thing. Presumably the success message would be around log level `info` or `debug`, and the warning message would be at log level `warn`.

dmjio:
@tysonzero I agree there should be additional granularity when it comes to prerendering and logging. I probably should not have added logging in this PR to begin with, since this PR now conflates a few things, but it was strategic for my own debugging. I think we should open a new issue related to logging granularity to fulfill this request, and potentially logging events as well cc @niteria

Issue #596, opened 2020-06-16
style element


niteria:
**Is your feature request related to a problem? Please describe.**
Now that I can run isomorphic apps with `jsaddle`, it'd be convenient for me to be able to add a style tag in the file that generates HTML that's used by both the client and the server.

I thought `<style>` inside `<body>` is disallowed, but apparently the world has moved on and it's part of HTML 5.2: https://stackoverflow.com/a/50154225

I have a stopgap solution like this:
```
nodeHtml "style" [] [text $ Miso.ms $ TL.toStrict $ Clay.renderWith Clay.pretty [] css]
```
but I'm worried that I might run into escaping issues by using `text`.

**Describe the solution you'd like**
A new function, say called `styleTag_` with the type `styleTag_ :: [Attribute action] -> MisoString -> View action` that adds a `<style>` tag.

dmjio:
@niteria, Sure, make a PR I'll accept, can we name it `style_` though?

niteria:
I thought there was a name collision with some other `style_`, but maybe there isn't. I will give it a try.

dmjio:
`style_` is an attribute, so it won't conflict, but you'll have to hide the element `style_` when re-exporting, otherwise the symbols will collide.

dmjio:
to use it you'd have to do
```haskell
import qualified Miso.Html as Html

view = Html.style_ [] "body { background : red; }"

**Comment by niteria** [comment]
Ok, I confirmed my concerns with escaping problems in my stopgap solution.

If I use:
```
nodeHtml "style" [] [text "li > a { text-decoration-color: red; }"]
```
The server will render that as:
```
<style>li &gt; a { text-decoration-color: red; }</style>
```

And the browser doesn't understand that.

Lucid (outside `miso`) deals with that by using `TermRaw` instead of `Term`: https://hackage.haskell.org/package/lucid-2.9.12/docs/src/Lucid.Html5.html#style_ for the `style_` element. The effect is that Lucid doesn't try to escape in `<style>` tags.

Currently, the `VTree` only has `VNode` and `VText` constructors. Text put in `VText` is always escaped. Right now I don't see another path than extending `VTree` with `VTextRaw`.

Does that seem reasonable?

niteria:
It sounds pretty invasive, maybe `style_` should be explicitly not supported. 

There's probably a way to tell `jsaddle` to attach some styles, it would solve my original problem. I will have a look.

niteria:
[...]

PR #601, opened 2020-06-24
Add style_ and script_

niteria:
This addresses #596

Test Plan: Run https://github.com/niteria/miso-isomorphic-example/tree/vtextraw
which uses style_ and script_.
The browser console prints "Successfully prendered page" with no errors.
Going between / and /flipped successfully toggles the color of the buttons.

### DIFF
--- a/ghc-src/Miso/Html/Internal.hs
+++ b/ghc-src/Miso/Html/Internal.hs
@@ -30,6 +30,7 @@ module Miso.Html.Internal (
   -- * Smart `View` constructors
   , node
   , text
+  , textRaw
   -- * Key patch internals
   , Key    (..)
   , ToKey  (..)
@@ -74,6 +75,14 @@ data VTree action where
            } -> VTree action
   VText :: { vText :: Text -- ^ TextNode content
            } -> VTree action
+  -- Invariant: To avoid complexity with collapsing mixed VText and
+  -- VTextRaw nodes, VTextRaw node is always the only child.
+  -- That's not a big limitation, since the intended purpose is to be able
+  -- to use <style> and <script> tags. That means we can't support HTML
+  -- entities like &nbsp; with VTextRaw, but the same can be achieved with
+  -- Unicode.
+  VTextRaw :: { vText :: Text -- ^ Raw TextNode content
+              } -> VTree action
   deriving Functor
 
 instance Show (VTree action) where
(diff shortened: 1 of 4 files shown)


src/Miso/Html/Element.hs
```diff
@@ -51,7 +51,9 @@ module Miso.Html.Element
     , canvas_
     , math_
     , script_
```
niteria, line 53: `script_` is dangerous, even something as simple as:
```
script_ [] [text "console.log('<test>')"]
```

Fails with:
```
SyntaxError: expected expression, got '&' localhost:3003:1:12
VTree differed from node Object { type: "vtext", text: "console.log('<test>')" } #text "console.log(&#39;&lt;test&gt;&#39;)" all.js:695:152
Could not copy DOM into virtual DOM, falling back to diff all.js:694:395
<test> line 685 > scriptElement:1:9
```

I'm tempted to reclaim the name `script_` for the new version that I'm adding. What do you think?

ghc-src/Miso/Html/Internal.hs
```diff
@@ -84,6 +93,8 @@ instance L.ToHtml (VTree action) where
   toHtmlRaw = L.toHtml
   toHtml (VText x) | T.null x = L.toHtml (" " :: MisoString)
                    | otherwise = L.toHtml x
+  toHtml (VTextRaw x) | T.null x = L.toHtml (" " :: MisoString)
```
niteria, line 96: I parroted the lines above, but I'm not sure if it's necessary.

src/Miso/Html/Element.hs
dmjio: I'd say go for it.

ghc-src/Miso/Html/Internal.hs
dmjio: That's fine, this show is just for debugging purposes anyways.

src/Miso/Html/Element.hs
niteria: Then I'll have `style_` for consistency as well.

niteria:
I just realized that `textRaw` gets re-exported in `Miso.Html`, I think if someone has a good reason to use it they should import the Internal module.

src/Miso/Html.hs
```diff
@@ -34,7 +34,7 @@ module Miso.Html
    , module Miso.Html.Property
    ) where
 
-import Miso.Html.Element
+import Miso.Html.Element hiding (style_)
 import Miso.Html.Event
-import Miso.Html.Internal
+import Miso.Html.Internal hiding (textRaw)
```
dmjio, line 39: 👍 nice, exactly what I was going to propose.

dmjio:
Tested out your branch, very impressive. Nice how CSS in body can be altered via style tags.

niteria:
Ok, I'm almost happy with this, I just need to add some warning about potential JS and CSS injection if not used carefully.

niteria:
Ok, I think it's ready. Thanks for the responding so quickly :)

dmjio:
@niteria Thanks for your great work !

dmjio:
Love the OWASP links as well.

dmjio:
@niteria you should consider joining the `miso` slack. There is a link on the README (one of the badges).

PR #604, opened 2020-06-28
Improve haddock for style_ and script_

niteria:
I didn't consider how haddock would render this when I first
committed and it's really ugly. This fixes that.


niteria:
Old:
Screenshot_20200628_230754

New:
Screenshot_20200628_230838

Issue #608, opened 2020-07-06
Add static css and javascript to output html and jsaddle


developandplay:
**Is your feature request related to a problem? Please describe.**
I'm currently loading css and javascript dynamically within Miso. However this approach does not work for Material Web Components that need to load javascript before Miso.

**Describe the solution you'd like**
A way to specify css and javascript files such that they will be static in the generated index.html or jsaddle.

**Describe alternatives you've considered**
I can edit the generated index.html to include the static css and javascript files or create my own html file. Unfortunately this is tedious and not possible when using jsaddle.

**Additional context**
I'm porting Material Web Components from Elm to Miso.

dmjio:
There is https://hackage.haskell.org/package/jsaddle-warp-0.9.7.0/docs/src/Language.Javascript.JSaddle.WebSockets.html#jsaddleAppWithJs

developandplay:
Thank you! As a beginner this now took me quite some time to figure out.
Is this something we could build a sample file for or add to the readme of `sample-app-jsaddle`?

In case anyone else stumbles across this problem:
First download the external javascript file (`your-file.js`) to your project directory.
Then add `bytestring` to `build-depends` in `app.cabal`.
In your `Main.hs`  you need to change the implementation of `runApp` from this:
```
runApp f =
  Warp.runSettings (Warp.setPort 8080 (Warp.setTimeout 3600 Warp.defaultSettings)) =<<
    JSaddle.jsaddleOr defaultConnectionOptions (f >> syncPoint) JSaddle.jsaddleApp
```
to this:
```
runApp f = do
  bString <- B.readFile "your-file.js"
  jSaddle <- JSaddle.jsaddleOr defaultConnectionOptions (f >> syncPoint) (JSaddle.jsaddleAppWithJs (B.append (JSaddle.jsaddleJs False) bString))
  Warp.runSettings (Warp.setPort 8081 (Warp.setTimeout 3600 Warp.defaultSettings)) jSaddle
```
Now you should be able to use `your-file.js` in jsaddle.

dmjio:
We could add this to the documentation here:
https://github.com/dmjio/miso/blob/master/sample-app-jsaddle/README.md

PR #610, opened 2020-07-07
Update README.md to add external javascript files

developandplay:
#608


dmjio:
Nice work, thank you.

Issue #631, opened 2020-10-23
JSaddle breaks `preventDefault` on links


tysonzero:
When building with the jsaddle flag enabled, links that use an `href` tag plus `preventDefault` cause a full page reload.

To reproduce apply this diff:

```patch
diff --git a/sample-app-jsaddle/Main.hs b/sample-app-jsaddle/Main.hs
index 96f765f..06fc710 100644
--- a/sample-app-jsaddle/Main.hs
+++ b/sample-app-jsaddle/Main.hs
@@ -66,4 +66,12 @@ viewModel x = div_ [] [
    button_ [ onClick AddOne ] [ text "+" ]
  , text (ms x)
  , button_ [ onClick SubtractOne ] [ text "-" ]
+ , a_ [ onClickPreventDefault AddOne, href_ "/bad" ] [ text "add" ]
  ]
+
+onClickPreventDefault :: a -> Attribute a
+onClickPreventDefault a = onWithOptions
+    (defaultOptions { preventDefault = True })
+    "click"
+    emptyDecoder
+    (\() -> a)
```

Then follow the instructions in the `sample-app-jsaddle` README. When running `release` (no jsaddle), everything works correctly. However when running `dev`, the `add` link causes a full page reload.

From experimentation on other projects it appears as though this issue remains even if compiling with `ghcjs` + `jsaddle`. Although the above reproduction is insufficient to demonstrate this.

tysonzero:
@cocreature since you developed the `jsaddle` flag, this seems like it might be in your wheelhouse.

PR #632, opened 2020-10-23
Change Delegate Event to use Synchronous Function

parthshah31:
Async function breaks event prevent default and more in jsaddle mode. This changes the async function to a synchronous function.


tysonzero:
Fixes #631

PR #636, opened 2020-12-08
Fixed Bulma nav for haskell-miso.org mobile

tjlaxs:
Accidentally noticed that https://haskell-miso.org/ Bulma nav was broken in mobile. Just added a div around the a-elements and moved classes to the div and now it works nicely. Tried to match the layouting with the original, but not sure what is happening in there.


dmjio:
@tjlaxs awesome ! thanks.

Issue #644, opened 2021-03-16
Diffing algorithm - newProp with value of zero leads to prop deletion


developandplay:
**Describe the bug**
If I'm using anything selection based like tabs or options which has a node with a zero index that node will get deleted.
I think the reason lies in line 96.
https://github.com/dmjio/miso/blob/69f50b49adbff8217f8b51ae5f47727ee950f204/jsbits/diff.js#L94-L96
If `newProp = 0` holds then this check will fail however zero is a valid value for a new prop.
I think `newProp === undefined` is probably the check that comes sematically the closest without deleting props with zero value.

**To Reproduce**
Either run [https://github.com/developandplay/material-components-web-miso](https://github.com/developandplay/material-components-web-miso) locally or check out our demo website [matiso.dev](http://www.matiso.dev.s3-website.eu-central-1.amazonaws.com/). Then select the first option as shown in the screenshots below and look at the error in the console.

**Expected behavior**
If a `newProp = 0` occurs it should be treated as a prop with valid content.

**Screenshots**
Expected behaviour when selecting the second option. `node[c] = newProp;` gets executed.
![image](https://user-images.githubusercontent.com/34752929/111390162-8e9ebb00-86b2-11eb-9712-174c6e125d96.png)

Unexpected behaviour when selecting the first option. `node[c] = '';` gets executed.
![image](https://user-images.githubusercontent.com/34752929/111390728-8430f100-86b3-11eb-9b6b-33663acff17a.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Chrome

dmjio:
Very nice catch.

PR #645, opened 2021-03-16
Diffing algorithm - allow props with zero value

developandplay:
Change proposed in #644


dmjio:
Well done. Nice find.

Issue #646, opened 2021-03-17
Dynamic properties ignored by decoder


developandplay:
**Describe the bug**
When the object that should be decoded has dynamic properties (getters) these will be ignored in the decoder.

**To Reproduce**
Either run [https://github.com/developandplay/material-components-web-miso](https://github.com/developandplay/material-components-web-miso) locally or check out our demo website [matiso.dev](http://www.matiso.dev.s3-website.eu-central-1.amazonaws.com/). Then move the slider and look at the error in the console. (The slider position coordinates might be off when dragging without resizing the screen.)

**Expected behavior**
Dynamic properties should be decoded just like static properties.

**Screenshots**
Unexpected behaviour with expections thrown if dynamic properties are not decoded. As seen in the second screenshot `value` is missing in `newObj` and only the static `value_` is included. In the third screenshot we can see that `value` is an element of `obj` but is a dynamic getter.
![image](https://user-images.githubusercontent.com/34752929/111553685-2ff64180-8785-11eb-9981-a9e5d98d0498.png)
![image](https://user-images.githubusercontent.com/34752929/111553897-8ebbbb00-8785-11eb-9634-51e453cf2451.png)
![image](https://user-images.githubusercontent.com/34752929/111554119-fffb6e00-8785-11eb-9aa9-5474dd25a7d4.png)

Expected behaviour with no exceptions thrown after including dynamic properties in e84d463aaed764810f8d16c033369e9b250a38f8. As seen in the last screenshot `value` is now added to `newObj` when previously only the static `value_` was included.
![image](https://user-images.githubusercontent.com/34752929/111552803-5dda8680-8783-11eb-9412-08043240c9b2.png)
![image](https://user-images.githubusercontent.com/34752929/111554362-81530080-8786-11eb-870f-20cccd484ea7.png)

**Desktop (please complete the following information):**
 - OS: ubuntu
 - Browser: chrome

dmjio:
Can we try doing this w/o the Set.

developandplay:
If you are thinking of replacing the Set with a regular Array then I don't think so.
The Set in this case insures that an attribute/method of an object lower down the protype chain actually [overshadows](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain#inheriting_methods) the ones higher up in the prototype chain.
So if we were to use an array we would still need to check if the attribute already exists.

That being said there might be completly different approaches to this problem that I didn't think of.

[...]

PR #647, opened 2021-03-17
Add dynamic properties to decoder

developandplay:
As mentioned in #646 
The additional `getAllPropertyNames` method doesn't look that pretty in my eyes but apparently no such method exists in standard javascript.

### DIFF
--- a/jsbits/delegate.js
+++ b/jsbits/delegate.js
@@ -116,7 +116,7 @@ window['objectToJSON'] = function objectToJSON (at, obj) {
   /* If obj is a non-list-like object */
   newObj = {};
   var isInput = obj['localName'] === 'input';
-  for (var i in obj){
+  for (var i of getAllPropertyNames(obj)){
     /* bug in safari, throws TypeError if the following fields are referenced on a checkbox */
     /* https://stackoverflow.com/a/25569117/453261 */
     /* https://html.spec.whatwg.org/multipage/input.html#do-not-apply */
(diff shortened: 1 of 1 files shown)


dmjio:
The delegator is some performance sensitive code, so I'm slightly worried about the Set allocation.

developandplay:
True now if javascript code was aware which prop of the object should be decoded that could be a more performant approach.
However I guess you probably put the actual decoding in Haskell and not on the javascript for a good reason.

dmjio:
@developandplay yea, decoding is in Haskell so people can build their own decoders with `aeson` (or other).

dmjio:
@developandplay been doing some benchmarks, seems adding this would only result in a +/- 0.1ms penalty. In some cases it is actually faster than the existing code... which is surprising. 

https://jsfiddle.net/61sw9de5/ 

^ you basically move your mouse (or click the button) and it benchmarks the time it takes to accumulate all properties w/ both versions of the code, times them, then displays the improvement (positive or negative).

developandplay:
Wow this is super cool. Thanks for revisiting the issue!

dmjio:
No problem, and thanks for this patch ;) 

It does seem that `new Set()` isn't supported in older versions of IE ...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#browser_compatibility

To get around this we could check the `document.documentMode` property.

https://www.w3schools.com/jsref/prop_doc_documentmode.asp

dmjio:
Or we can remove the Set, and just use a regular object.

```javascript
function getAllPropertyNames(obj) {                                                                                                                                                                                                 
    var props = {}; 
    ... 
    return Object.keys(props);                                                                                                                                                                                                         
}   
```

developandplay:
Sounds like a decent alternative.
Seems like that was the way to go when set was not a thing yet.

I would expect overshadowing to work as expected.
However it's probably better to double check that since it's the only reason we were using set in the first place.

Also let's add a comment then that we are only using it to "simulate" a set.
Now I'm curious to see if the performance is any different but I would guess no.

dmjio:
Merged in https://github.com/dmjio/miso/commit/b32d796ff6ca4434b986fae7844b1df8fe711b54

Issue #650, opened 2021-04-06
Broken links to Mozilla docs in Miso.Html.Property


niteria:
**Describe the bug**
It seems that Mozilla no longer hosts links such as https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/autoplay

https://github.com/dmjio/miso/blob/30e4844cc94a97751b28fcaee2004d56d5cfde51/src/Miso/Html/Property.hs#L297 is an example place where that's referenced.

dmjio:
https://github.com/dmjio/miso/commit/615d642619b6b19627fa0cf346bfe2484d4a6d6e

should fix this

PR #651, opened 2021-04-06
Add loading_ property

niteria:
According to https://www.w3schools.com/tags/att_img_loading.asp
it can be used to make the images lazily load.


niteria:
Docs point nowhere as per #650.

Issue #597, opened 2020-06-18
sample-app-jsaddle doesn't work under Firefox


niteria:
**Describe the bug**
If you build the `sample-app-jsaddle` app and open it in `Firefox` the app will hang, loop sending many XHR requests and you get `Error : Unexpected Duplicate. syncCallbacks=True nBatch=4 nExpected=6` on `stdout`.

This is not really an issue with `miso`, but with `jsaddle`. `miso` is triggering https://github.com/ghcjs/jsaddle/issues/64

I'm opening this because I think this is easy to side-step in `miso`.

**Additional context**
The crux of the problem with `jsaddle` + `Firefox` is that `Firefox` (as opposed to `Chrome`) is strict about disallowing synchronous XHR requests.

Synchronous XHR requests are used in `jsaddle` to implement calls from the browser to the `jsaddle` backend for Haskell functions that return results.

Fortunately `miso` is written in a continuation-passing style, so as far as I can tell it has no need for such functions.

In fact, the only instance of using a sync function that I run into is in https://github.com/dmjio/miso/blob/af3363397994e855760d646fc5deb9718dcd550b/jsaddle-ffi/Miso/FFI.hs#L221-L227.

It's clearly not using the result, and changing from `function` to `asyncFunction` on this line https://github.com/dmjio/miso/blob/af3363397994e855760d646fc5deb9718dcd550b/jsaddle-ffi/Miso/FFI.hs#L223 made the `sample-app-jsaddle` and my own app run successfully under `jsaddle` on `Firefox`.

niteria:
Fixed by afe2536b08c7bea5267b1e7b0121c786209dcd77

Issue #668, opened 2021-09-14
miso jsaddle is broken with 1.8 (at least in firefox)


saschaheylik:
**Describe the bug**
After upgrading to 1.8, my application no longer works when trying to run with jsaddle. The GHCJS build still works. You can reproduce the bug using sample-app-jsaddle.

The bug seems to occur in Firefox but not in Chrome. Sometimes another Error is shown: "Error : Unexpected Duplicate. syncCallbacks=True nBatch=4 nExpected=6"

**To Reproduce**
1. Go to miso/sample-app-jsaddle
2. Change the tarball in the default.nix url to "https://github.com/dmjio/miso/archive/refs/tags/1.8.tar.gz".
2. run nix-shell --run reload
3. Browse to localhost:8080 to try to use the app

**Results:**
1. The app is unresponsive
2. This error is displayed in the terminal:
jsaddle missing sync message handler
CallStack (from HasCallStack):
  error, called at src/Language/Javascript/JSaddle/WebSockets.hs:129:40 in jsaddle-warp-0.9.7.0-DHMZWfEwF6AJVIaTVkQdus:Language.Javascript.JSaddle.WebSockets

saschaheylik:
The bug seems to occur in Firefox but not in Chrome. Sometimes another Error is shown: Error : Unexpected Duplicate. syncCallbacks=True nBatch=4 nExpected=6

dmjio:
![image](https://user-images.githubusercontent.com/875324/134032837-9c6a1ea3-aaa9-4641-b82c-df3b76a8f34c.png)

While I do see the app does become unresponsive, I'm unable to reproduce the exact error you're experiencing in the console.

```
CallStack (from HasCallStack):
error, called at src/Language/Javascript/JSaddle/WebSockets.hs:129:40 in jsaddle-warp-0.9.7.0-DHMZWfEwF6AJVIaTVkQdus:Language.Javascript.JSaddle.WebSockets
```

dmjio:
```haskell
f x =                                                                                                                                                                                         
  onWithOptions defaultOptions { preventDefault = True }                                                                                                                                      
    "click" emptyDecoder (\() -> x)                                                                                                                                                           
                                                                                                                                                                                              
                                                                                                                                                                                              
-- | Constructs a virtual DOM from a model                                                                                                                                                    
viewModel :: Model -> View Action                                                                                                                                                             
viewModel x = div_ [] [                                                                                                                                                                       
[...]

PR #669, opened 2021-09-17
Revert "Change Delegate Event to use Synchronous Function"

cdfa:
This causes Firefox to spin indefinitely when using JSaddle.
Context: https://github.com/dmjio/miso/issues/597

Not sure why this was introduced again. Handling `e.preventDefault` or `e.stopPropagation` purely in JS would be best, but that's going to take more work than this fix.

### DIFF
--- a/src/Miso/FFI.hs
+++ b/src/Miso/FFI.hs
@@ -240,7 +240,12 @@ jsStringToDouble = read . unpack
 -- | Initialize event delegation from a mount point.
 delegateEvent :: JSVal -> JSVal -> JSM JSVal -> JSM ()
 delegateEvent mountPoint events getVTree = do
+-- using asyncfunction on GHC causes an XHR request loop with JSaddle
+#ifdef __GHCJS__
+  cb' <- asyncFunction $ \_ _ [continuation] -> do
+#else
   cb' <- function $ \_ _ [continuation] -> do
+#endif
     res <- getVTree
     _ <- call continuation global res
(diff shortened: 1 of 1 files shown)


dmjio:
It is handled in JS https://github.com/dmjio/miso/blob/master/jsbits/delegate.js#L62

All events in miso are processed asynchronously, this doesn't jive well with `jsaddle` apparently, since it has to roundtrip to the server. I typically only use GHCJS to compile + develop.

We'll most likely need this behavior enabled for GHCJS, but disabled for GHC jsaddle.


```haskell
#ifdef __GHCJS__
cb' <- asyncFunction $ \_ _ [continuation] -> do
#else
cb' <- function $ \_ _ [continuation] -> do
#endif
  ```

dmjio:
So, upon second inspection. This code, when compiled w/ GHCJS, will convert `syncCallbacks` to `asyncCallbacks`, since it is using `ContinueAsync`. Not so the case with GHC + jsaddle. I've reverted to using `asyncFunction` for the `jsaddle` parts, but am testing to ensure this doesn't break `preventDefault` regarding links.

dmjio:
Upon second inspection, while all events in `miso` are handled asynchronously, calling into Haskell land should use synchronous callbacks (which is related to the code in question here). Otherwise, this causes miso to handle `preventDefault` improperly. For more info please see this https://github.com/dmjio/miso/issues/668#issuecomment-924433573

PR #672, opened 2021-09-22
Remove CPP related to an iOS build.

dmjio:


dmjio:
https://github.com/dmjio/miso/pull/767

PR #680, opened 2021-10-12
Trim miso deps when built with ghcjs

tysonzero:
Also cleaned up cabal file a bit in the process


tysonzero:
Not sure what's going on with the test matrix. `nix-build` works just fine locally.

PR #708, opened 2022-08-22
Fix GHC build with `text >=1.2.5`

amesgen:
Without this patch, we have
```
 $ cabal build --constraint='text >=1.2.5'
src/Miso/FFI.hs:95:35: error:
    Ambiguous occurrence ‘elem’
    It could refer to
       either ‘Prelude.elem’,
              imported from ‘Prelude’ at src/Miso/FFI.hs:14:8-15
              (and originally defined in ‘Data.Foldable’)
           or ‘Miso.String.elem’,
              imported from ‘Miso.String’ at src/Miso/FFI.hs:67:1-28
              (and originally defined in ‘Data.Text’)
   |
95 |   classSet <- ((JSS.pack "class") `elem`) <$> listProps obj
   |                                   ^^^^^^
```
where `elem` [was added](https://hackage.haskell.org/package/text-2.0.1/changelog) to `Data.Text` in 1.2.5.0.

### DIFF
--- a/src/Miso/FFI.hs
+++ b/src/Miso/FFI.hs
@@ -64,7 +64,7 @@ import           Language.Javascript.JSaddle hiding (obj, val)
 #else
 import           Language.Javascript.JSaddle hiding (Success, obj, val)
 #endif
-import           Miso.String
+import           Miso.String hiding (elem)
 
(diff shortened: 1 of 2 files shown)


dmjio:
@amesgen thanks

PR #730, opened 2023-09-18
Fix simple example source link

Andriamanitra:
The links to SSE and Pixel Card Wars also seem to be dead but I'm not sure if that's temporary so I left them be


dmjio:
🕋

PR #740, opened 2024-05-03
Update examples to use new JS backend syntax

sevanspowell:
- Three miso examples (canvas2d, file-reader, and threejs) make use of `foreign import`s. These need to be updated to the new GHC JS backend syntax in order to work with the new GHC JS backend.
  - I'm not sure these updates are backwards compatible, so I might have to conditionally introduce these changes (@hamishmack ?). 
- Additionally https://github.com/ghcjs/ghcjs-base/pull/136 must be merged to fix issues with ghcjs-base introduced in https://github.com/ghcjs/ghcjs-base/pull/135. The examples will fail to run otherwise.

### DIFF
--- a/examples/canvas2d/Main.hs
+++ b/examples/canvas2d/Main.hs
@@ -70,27 +70,27 @@ updateModel (sun,moon,earth) (SetTime m@(secs,millis)) _ = m <# do
   drawImage sun 0 0 300 300 ctx
   pure GetTime
 
-foreign import javascript unsafe "$1.globalCompositeOperation = 'destination-over';"
+foreign import javascript unsafe "((x) => { x.globalCompositeOperation = 'destination-over'; })"
   setGlobalCompositeOperation :: Context -> IO ()
 
-foreign import javascript unsafe "$4.drawImage($1,$2,$3);"
+foreign import javascript unsafe "((x, y, z, w) => { w.drawImage(x,y,z); })"
   drawImage' :: Image -> Double -> Double -> Context -> IO ()
 
-foreign import javascript unsafe "$r = document.getElementById('canvas').getContext('2d');"
+foreign import javascript unsafe "(() => { return document.getElementById('canvas').getContext('2d'); })"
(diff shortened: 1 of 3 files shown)


dmjio:
Thanks for doing this, in order to ensure it works we'll need to update miso's nix scripts to use the new ghcjs.

@stepcut has pointed out that `nix-shell -p 'pkgsCross.ghcjs.haskell.packages.ghcHEAD.ghcWithPackages (p: [ ])'` is working now on `nixpkgs-unstable`, but it seems a lot of dependencies are still broken

sevanspowell:
Thanks for your work on the project!

I'll do my best to help out with that process.

alexfmpe:
Can confirm this fixes build failures: https://github.com/NixOS/nixpkgs/pull/380737
Ran into
```
Building executable 'file-reader' for miso-examples-1.8.5.0...
[1 of 1] Compiling Main             ( file-reader/Main.hs, dist/build/file-reader/file-reader-tmp/Main.o )
file-reader/Main.hs:15:1: error: [8;;https://errors.haskell.org/messages/GHC-61948GHC-619488;;]
    Could not find module ‘GHCJS.Foreign.Callback’.
    Perhaps you meant GHC.JS.Foreign.Callback (from base-4.21.0.0)
    Use -v to see a list of the files searched for.
   |
15 | import GHCJS.Foreign.Callback
```
and then found this PR

alexfmpe:
Probably want to do a release with this and #752

dmjio:
@alexfmpe @sevanspowell 

```diff
- import GHCJS.Foreign.Callback
+ import GHC.JS.Foreign.Callback
```

Will break GHCJS builds, at the very least we should use CPP to make this a backwards compatible change.

dmjio:
Merged in #757 @alexfmpe @sevanspowell .

@alexfmpe I'll get a release out for https://github.com/NixOS/nixpkgs/pull/380737

PR #747, opened 2024-10-15
Update Pixel Card Wars' URL to the production one

smelc:
The URL of the Pixel Card Wars game was down (this was the staging URL). Change the link to use the production URL.

# How to trust this PR

Click on the new link, witness it works.


dmjio:
Looks great

PR #751, opened 2024-10-31
Expose global sink

dmjio:
@functora


dmjio:
cc @tysonzero

dmjio:
Could even do

```haskell
globalSink :: Sink action
globalSink = unsafePerformIO (readIORef globalSinkRef) 
```

PR #752, opened 2024-11-01
Fix build with `text` 2.1.2

georgefst:
https://hackage.haskell.org/package/text-2.1.2/changelog

We need to deal with `show` now being exported from `Data.Text`.

Strictly speaking, all old versions of Miso need a Hackage revision setting `text < 2.1.2`.

### DIFF
--- a/src/Miso/FFI.hs
+++ b/src/Miso/FFI.hs
@@ -64,7 +64,7 @@ import           Language.Javascript.JSaddle hiding (obj, val)
 #else
 import           Language.Javascript.JSaddle hiding (Success, obj, val)
 #endif
-import           Miso.String
+import           Miso.String hiding (show)
 
 -- | Run given `JSM` action asynchronously, in a separate thread.
 forkJSM :: JSM () -> JSM ()
--- a/src/Miso/Html/Types.hs
+++ b/src/Miso/Html/Types.hs
@@ -65,7 +65,7 @@ import           Text.HTML.TagSoup          (Tag(..))
 import           Miso.Effect
 import           Miso.Event
 import           Miso.FFI
-import           Miso.String                hiding (reverse, elem)
+import           Miso.String                hiding (reverse, elem, show )
 
 -- | Core type for constructing a `VTree`, use this instead of `VTree` directly.
 data View action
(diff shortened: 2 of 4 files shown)


georgefst:
@dmjio Could I nudge you for a release and revisions here? I keep having to vendor Miso in all of my projects because of the lax bounds.

dmjio:
Coming back to this, I think the better solution would be to not re-export `show` from `Miso.String`, as opposed to hiding explicitly in all imports. `text` also ships as a boot package w/ GHC and its not overrideable w/ nix haskell builds.

```haskell
{-# LANGUAGE CPP #-}

module Miso.String ( module X ) where

#if MIN_VERSION_text(2,1,2)
import qualified Data.Text as X hiding (show)
#else
import qualified Data.Text as X
#endif 
```

georgefst:
> Coming back to this, I think the better solution would be to not re-export `show` from `Miso.String`, as opposed to hiding explicitly in all imports.

Yes, much better idea! I can't remember why I did it my way to be honest.

I will say again, Hackage revisions for old versions would be useful in order to prevent bad build plans. This is quite easy to do with [hackage-cli](https://hackage.haskell.org/package/hackage-cli).

dmjio:
It should be fine now that the CPP is in. Regarding build plans, they aren't necessary if nix is used as your resolver. Nix is unfortunately still necessary for reasonably working with GHC cross compilers. It also keeps us from playing "package bumper" all the time. The external APIs for most of miso's deps. don't change, like `text`. Nobody is changing `pack` for example.

georgefst:
> Nix is unfortunately still necessary for reasonably working with GHC cross compilers.

This isn't really true for the Wasm backend. And even when using Nix there, the common approach is to use the `ghc-wasm-meta` flake, which means using Nix to get hold of an appropriately compiled and wrapped GHC and Cabal, but _not_ using Nix as the resolver.

Anyway, I can't force you to do it! And I'm aware that there's some PvP skepticism among Nix and Stack users. I will just have to keep a mental note that not using the latest version of Miso might result in build failures.

dmjio:
So far there's been no build failures reported by users until `text` (mostly heavily used package in the ecosystem) changed its external API. Ideally we can integrate GHC WASM into nixpkgs correctly so that it can take advantage of cached packages. Not sure this is feasible thought because it uses a non-upstreamable fork of LLVM.

dmjio:
If it starts getting out of hand I'll add constraints.

PR #756, opened 2025-02-09
Hide `show` re-exports on text > 2.1.2 in Miso.String

dmjio:
Hides external API changes (`show` re-export) for `Data.Text`


dmjio:
Related to https://github.com/dmjio/miso/pull/752

PR #757, opened 2025-02-09
New syntax support, backwards compat cpp for ghcjs

dmjio:
Brings in https://github.com/dmjio/miso/pull/740, adds CPP for ghcjs.

### DIFF
--- a/examples/file-reader/Main.hs
+++ b/examples/file-reader/Main.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -12,7 +13,11 @@ import           Miso.String
 import           Control.Concurrent.MVar
 
 import GHCJS.Types
+#ifndef ghcjs_HOST_OS
 import GHC.JS.Foreign.Callback
+#else
+import GHCJS.Foreign.Callback
+#endif
 
 -- | Model
 data Model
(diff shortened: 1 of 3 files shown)


examples/file-reader/Main.hs
```diff
@@ -71,6 +76,22 @@ viewModel Model {..} = view
       , div_ [] [ text info ]
       ]
 
+#ifndef ghcjs_HOST_OS
```
alexfmpe, line 79: @dmjio 
This is defined both for ghcjs and ghc's js backend. 
You want something like this
https://github.com/ghcjs/jsaddle/blob/b72b4d01fdc186d2c7447beb7273ad13358ff593/jsaddle/src/Language/Javascript/JSaddle/Exception.hs#L21-L27

dmjio: https://github.com/ghc/ghc/blob/master/testsuite/tests/javascript/T23346.hs

alexfmpe: > @alexfmpe where do you see ghcjs_HOST_OS defined for the new JS backend?

Hmm? AFAICT cabal considers it `os(ghcjs)` for both old ghcjs and new js backend https://hackage.haskell.org/package/Cabal-syntax-3.14.1.0/docs/Distribution-System.html#v:Ghcjs 
unlike `impl(ghcjs)`
https://hackage.haskell.org/package/Cabal-syntax-3.14.1.0/docs/Distribution-Compiler.html#v:GHCJS

dmjio: that's.... less than ideal, but found a workaround https://github.com/dmjio/miso/issues/759

PR #769, opened 2025-02-20
Fix type application parse error on GHC 9+

georgefst:
See https://downloads.haskell.org/~ghc/9.0-latest/docs/html/users_guide/9.0.1-notes.html.


georgefst:
Note that this is enough for `cabal build all -w ghc-9.10` to be successful.

PR #774, opened 2025-02-21
 Use environment variable for runner port 

georgefst:
First two commits clean things up a bit in passing.

### DIFF
--- a/miso.cabal
+++ b/miso.cabal
@@ -30,7 +30,7 @@ common jsaddle-selector
   if arch(wasm32)
     build-depends:
       jsaddle-wasm
-  else
+  elif !impl(ghcjs)
     build-depends:
       jsaddle-warp
 
--- a/miso.cabal
+++ b/miso.cabal
@@ -30,7 +30,7 @@ common jsaddle-selector
   if arch(wasm32)
     build-depends:
       jsaddle-wasm
-  elif !impl(ghcjs)
+  else
     build-depends:
       jsaddle-warp
 
--- a/src/Miso/Runner.hs
+++ b/src/Miso/Runner.hs
@@ -19,7 +19,7 @@ run = J.run
 #elif defined(ghcjs_HOST_OS)
 run = id
 #else
-run x = getProgName >>= \case
-    "<interactive>" -> J.debug 8008 x
-    _ -> J.run 8008 x
+run action = do
+    isGhci <- (== "<interactive>") <$> getProgName
+    (if isGhci then J.debug else J.run) 8008 action
 #endif
--- a/src/Miso/Runner.hs
+++ b/src/Miso/Runner.hs
@@ -4,6 +4,7 @@ module Miso.Runner (run) where
 
 #if defined(wasm32_HOST_ARCH)
 import qualified Language.Javascript.JSaddle.Wasm as J
+#elif defined (ghcjs_HOST_OS)
 #else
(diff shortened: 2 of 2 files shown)


miso.cabal
```diff
@@ -30,7 +30,7 @@ common jsaddle-selector
   if arch(wasm32)
     build-depends:
       jsaddle-wasm
-  else
+  elif !impl(ghcjs)
```
dmjio, line 33: We should be able to keep this as `else`, since `jsaddle-warp` on js backends is just a no-op and introduces no new deps

```haskell
run :: Int -> JSM () - >IO ()
run _ m = m
```

src/Miso/Runner.hs
```diff
@@ -4,24 +4,25 @@ module Miso.Runner (run) where
 
 #if defined(wasm32_HOST_ARCH)
 import qualified Language.Javascript.JSaddle.Wasm as J
+#elif defined (ghcjs_HOST_OS)
```
dmjio, line 7: same as comment above, this should be unnecessary

src/Miso/Runner.hs
```diff
@@ -4,24 +4,25 @@ module Miso.Runner (run) where
...
 #endif
 
 import           Language.Javascript.JSaddle
 
 -- | Entry point for a miso application
-#if defined(wasm32_HOST_ARCH)
 run :: JSM () -> IO ()
+#if defined(wasm32_HOST_ARCH)
 run = J.run
+#elif defined(ghcjs_HOST_OS)
```
dmjio, line 21: Same as above. The `run` introduced by `jsaddle-warp` is identical here, so we can remove this `#elif` case as well and just use a hardcoded port that gets ignored on js backends.

```haskell

-- | Run the given 'JSM' action as the main entry point.  Either directly
--   in GHCJS or as a Warp server on the given port on GHC.
#ifdef ghcjs_HOST_OS
run :: Int -> IO () -> IO ()
run _port = id
#else
run :: Int -> JSM () -> IO ()
run port f =
    runSettings (setPort port (setTimeout 3600 defaultSettings)) =<<
        jsaddleOr defaultConnectionOptions (f >> syncPoint) jsaddleApp
#endif
```

miso.cabal
georgefst: I figured dropping the dependency completely was simpler, seeing as we're already bothering with Cabal conditionals and CPP anyway for the Wasm stuff, but fair enough, I can revert it

src/Miso/Runner.hs
```diff
@@ -4,24 +4,25 @@ module Miso.Runner (run) where
...
+run = id
 #else
-run :: JSM () -> IO ()
-#ifndef ghcjs_HOST_OS
-run x = getProgName >>= \case
-    "<interactive>" -> J.debug 8008 x
-    _ -> J.run 8008 x
-#else
-run = J.run 8008
```
dmjio, line 25: We can keep 

```haskell
#ifndef ghcjs_HOST_OS
run action = do
    port <- (readMaybe =<<) <$> lookupEnv "PORT"
    isGhci <- (== "<interactive>") <$> getProgName
    (if isGhci then J.debug else J.run) (fromMaybe 8008 port) action
#else
run = J.run 8008
#endif 
```

dmjio (changes requested):
See comments regarding CPP @georgefst

PR #777, opened 2025-02-22
Refactor NixOS modules, and NixOS test runner.

dmjio:
- Add NixOS test runner
  - Test the service can be curl'd.
- Separate out nixops specific (deployment objects) from other modules
- Add nginx.nix, server.nix -> machine.nix
- Don't rely on packageOverrides in this hash of nixpkgs


dmjio:
```
Configuring KVM
 Accelerated KVM is enabled ⚡️
 ```
 
 Use @DeterminateSystems nix installer

PR #780, opened 2025-02-26
Fix broken source link on haskell-miso.org

ashwinmathi:
Didn't want to make a whole issue for such a minor change but the recent move of the haskell-miso.org source to the top-level directory broke the website's link to its source code. This fixes the issue.


dmjio:
thank you

PR #793, opened 2025-03-12
Add upper bounds to miso.cabal file.

dmjio:
Now that miso is supported by the GHC WASM backend (which is integrated with the cabal workflow). We should begin specifying version bounds for cabal's resolver as opposed to relying on nixpkgs to pin all versions.

cc @georgefst @hvr


georgefst:
So Miso builds with `base-0.1`? Awesome.

Issue #452, opened 2018-08-24
`onClick` does not always work on mobile safari


tysonzero:
To reproduce make an app with an `onClick` attribute on something that isn't a button and without any special styling, and try and click on it on mobile safari. You will notice that the click is not registered by Miso.

This is due to [this](https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html).

It's an unfortunate quirk, but it seems to me that Miso might be the right place to handle this, to maximize its cross browser compatibility.

dmjio:
@tysonzero can we close this?

tysonzero:
It's still an issue actually, on mobile safari the only way to close our login / signup modal is to tap on the "X", whereas on every other browser we have tried you can just click outside the modal.

dmjio:
Sadly I don't think cross compiling to ARM will solve this, since it still uses the same webkit rendering engine.

dmjio:
This should be addressed by `PointerEvent` in https://github.com/dmjio/miso/pull/807

dmjio:
This is fixed in master from #807 by `onpointerdown` . Tested on iOS Safari and OSX Safari. 

To test, try the example at simple.haskell-miso.org

PR #799, opened 2025-03-13
Split out events.

dmjio:
For performance reasons its good if we don't listen on all mouse / drag events. It's best if users add these incrementally to their applications. This should reduce congestion in the event queue.

cc @georgefst @brakubraku

### DIFF
--- a/examples/todo-mvc/Main.hs
+++ b/examples/todo-mvc/Main.hs
@@ -89,7 +89,7 @@ main = run $ startApp App { initialAction = NoOp, ..}
     model      = emptyModel
     update     = updateModel
     view       = viewModel
-    events     = defaultEvents
+    events     = defaultEvents <> keyboardEvents
     mountPoint = Nothing
     subs       = []
     logLevel   = Off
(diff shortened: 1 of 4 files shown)


georgefst:
As I've said on Matrix, I've been frustrated by the current API being a bit arbitrary, so I'm in favour of this.

I do wonder whether we should add [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) while we're at it.

dmjio:
@georgefst `PointerEvent` sounds pretty amazing, since its a generalization over `"mouse*"`, `"touch*", "pen*" etc. How about we add that in a separate PR (`pointerEvents`) to get some usage, and then discuss if we should be defining our external API in terms of `PointerEvent` for everything else, and if so make it the default.

dmjio:
`PointerEvent` might also have implications for this issue @tysonzero https://github.com/dmjio/miso/issues/452

georgefst:
> How about we add that in a separate PR (`pointerEvents`) to get some usage, and then discuss if we should be defining our external API in terms of `PointerEvent` for everything else, and if so make it the default.

Sounds good.

PR #807, opened 2025-03-15
Initial commit of PointerEvent

dmjio:
- Introduces 'move', 'out', 'cancel', 'enter', 'down', etc. for pointer
- Defines PointerEvent type, pointerDecoder
- Move svg an simple examples to PointerEvent


dmjio:
@georgefst

PR #811, opened 2025-03-16
Allow Components to sample each other's local state.

dmjio:
- Allows child components to sample a parent component's state
- Saves each component's local state into an `IORef` that gets updated during the event loop post-diffing.
- If the component being sampled is not currently mounted a `Nothing` is returned.
- Adds example usage of `sample` to components
- Consider `MonadFail` instead of `Maybe`


dmjio:
@georgefst

PR #815, opened 2025-03-17
Intitial commit of DebugEvents

dmjio:
- Checks that all event handlers defined are being listened for by top-level event delegation (local to the Component mount or top-level App mount).

- Adds a warning if an event that was raised did not find an event handler for dispatch.

- Fix detected bug (this feature already found a bug in the svg example). Adds "pointermove" to the top-level Events map

- Introduce 'Events' type (synonym for Map MisoString Bool)

- Adds DebugEvent and DebugAll

- Renames some ambiguous variables to be more descriptive.

- Passes event Map from App{..} into Event construct continuation.

~~- Adds `componentEvents` to `ComponentState`~~

- Renames P,E,S to Property, Event, Style.

This detects if an event, that a user is expecting to fire, is not being listened on by the top level event delegator DOM node (typically body).

```haskell
main :: IO ()
main = run $ startApp app
   { events = pointerEvents
   }
 
 viewModel :: Model -> View Action
 viewModel x =
     div_
         []
         [ button_ [onPointerDown AddOne] [text "+"]
         , text (ms x)
         , button_ [onPointerDown SubtractOne] [text "-"]
         , button [ onClick NoOp ] [ text "oops" ] 
         -- ^ since `"click"` is not in `pointerEvents`, 
         -- an error will be raised in `console.error`
         ]

### DIFF
--- a/examples/svg/Main.hs
+++ b/examples/svg/Main.hs
@@ -18,7 +18,7 @@ foreign export javascript "hs_start" main :: IO ()
 
 main :: IO ()
 main = run $ startApp app
-  { events = pointerEvents
+  { events = M.insert "pointermove" False pointerEvents
   , subs = [ mouseSub HandlePointer ]
   }
 
--- a/haskell-miso.org/shared/Common.hs
+++ b/haskell-miso.org/shared/Common.hs
@@ -93,7 +93,7 @@ haskellMisoComponent ::
 haskellMisoComponent uri
   = component (app uri)
   { subs = [uriSub HandleURI]
-  , logLevel = DebugPrerender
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: examples/svg/Main.hs [hunk 1]
```diff
@@ -18,7 +18,7 @@ foreign export javascript "hs_start" main :: IO ()
 
 main :: IO ()
 main = run $ startApp app
-  { events = pointerEvents
+  { events = M.insert "pointermove" False pointerEvents
```
dmjio, line 21: We already detected a bug with this new debug feature.

PR #766, opened 2025-02-19
Initial commit of Components

dmjio:
🍜  Miso Components 
================================

Adds React-style components to Miso. This allows the embedding of `App` within `App` using an existential wrapper `SomeComponent` implemented as a new branch of the `View` tree.

⚠️ This is a work in progress ⚠️

### API

```haskell
data View action
  = Node NS  MisoString (Maybe Key)  [Attribute action]  [View action] 
  | Text MisoString (View action)                                                                                                                                                                                                       
  | TextRaw MisoString  (View action)                                                                                                                                                                                                    
  | ComponentNode SomeComponent (ComponentOptions action) (View action)                                                                                                                                                                                               
                                                                                                                                                                                                                                          
data SomeComponent = forall name model action . Eq model => SomeComponent (Component name model action)                                                                                                                                                           

data Component (name :: Symbol) model action
  = Component
   { name :: MisoString
   , app :: App model action
   }

data ComponentOptions action
  = Component
   { onMounted :: Maybe action
   , onUnmounted :: Maybe action
   , key :: Maybe Key
   , attributes :: [ Attribute action ]
   }

-- Smart constructor
componentOptions :: ComponentOptions action
componentOptions = Component Nothing Nothing Nothing []

component :: App model action -> Component name model action
component app = Component (symbolVal @name) app
-- ^ dmj: we do this so 'name' is statically known and used as the `data-component-id`
                                                                                                                                                                                                                                          
embed :: Eq model => Component name model a -> View action                                                                                                                                                                                       
embed app = SomeComponent (Component app) componentOptions -- used for mount / unmount hooks                                                                                                                                                                                                                                           
-- ^ note how `action` and `a` are different. This is how we embed `App` inside each other (recursively)

embedWith :: Eq model => Component name model a -> ComponentOptions action -> View action                                                                                                                                                                                       
embedWith app = SomeComponent (Component app)

-- example
calendar :: Component "calendar-app" CalendarModel CalendarAction
calendar = component calendarApp

view m = div_ [ ] [ embed calendar ]

notify :: Component name model action -> action -> Effect a m
-- ^ `actions`'s unify here, ensuring that we're sending the right `action` to the right `App`

-- example
update action model = do
  notify calendar (NewEntry (todo model))
 ```

- Adds `SomeComponent` existential wrapper used for nesting `App`
- Adds `notify` for bidirectional communication between components (parent, child or siblings)
- Modifies `diff.js` to mount / unmount components
[...]

PR #823, opened 2025-03-21
Bump nixpkgs hash to use 9.12.2 (new JS backend)

dmjio:
- [x] Uses new JS backend
- [x] Updates nix scripts to use `pkgsCross.ghcjs`
- [x] Addresses some warnings
- [x]  `ReferenceError: h$splitmix_init is not defined`
- [x]  `> cp: cannot stat 'dist/build/autogen/autogen.jsexe': No such file or directory`
- [ ] `nixops` updates
- [x] Update QEMU NixOS test runner script
- [ ] `ERROR - [JSC_LANGUAGE_FEATURE] This language feature is only supported for UNSTABLE mode or better: Public class fields.`

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -72,4 +72,7 @@ in with pkgs.haskell.lib;
 
   # misc. examples
   inherit (pkgs) more-examples;
+
+  # typescript bundler / minifer / builder
+  inherit (pkgs) bun;
 }
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -63,6 +63,8 @@ options: self: super: {
     };
   };
 
+  bun = (import <nixpkgs> {}).bun;
+
   nixops = super.nixops.overrideAttrs (drv: {
     src = builtins.fetchTarball {
       url = "https://nixos.org/releases/nixops/nixops-1.7/nixops-1.7.tar.bz2";


mastarija:
Hey, just got into playing with the new JS backend so I remembered that `miso` exists. Do you need any help with getting it to work with the new JS backend? I should be able to put in some work on both Haskell and Nix side if you could point me in the right direction.

dmjio:
Hi @mastarija. Thanks for asking, and sure.

Right now `miso` does seem to work with both the (latest) JS backend and the WASM backend. The hard part is acquiring the compilers. This has now gotten a lot easier with tools like [GHCUp](https://www.haskell.org/ghcup/) and [Flakes](https://nix.dev/concepts/flakes.html).

```bash
nix shell 'gitlab:haskell-wasm/ghc-wasm-meta?host=gitlab.haskell.org'
```

This PR is moreso about bringing `miso`'s nixpkgs into modernity (think our hash is from 2018) and `miso` still cuts its releases from GHCJS 8.6 (since that ime is most stable - this might no longer be the case with the new backend). So `miso` aside, a lot of other stuff we depend on has either changed dramatically or been dropped outright (`closure-compiler`, `NixOS` testing framewor's Perl DSL, `nixops` has also been dropped - will need to resurrect that). So that's what this PR is about mostly.

If you'd like to get involved on the edge of development there are some opportunities. 

One thing we have added recently that I would like to see more large-scale usage of is [React-style components](https://github.com/dmjio/miso/pull/766). This should open the door to integrating nicely with components from CSS-only frameworks like [Bulma](https://bulma.io), [Tailwind's DaisyUI component library](https://daisyui.com/) and [Flowbite](https://flowbite.com/). These websites have the component rendered side-by-side with the JS / TS code. I would like to have an example of the same, except the components are rendered side-by-side with the `miso` code. I think this will help spur adoption of Haskell in the broader community as well.

If you do this we could host it at one of our subdomains. This is what @ptigwe did for the [2048 example](2048.haskell-miso.org) and [flatris](flatris.haskell-miso.org)

This would also be a good demonstration of the `notify` (message passing between components) /  [sample](https://haddocks.haskell-miso.org/Miso.html#v:sample) functions (accessing other component states) that were added. You can find more examples of that in the `examples/components` sub folder (very crude though).

Also, feel free to join our [matrix.org server](https://matrix.to/#/#haskell-miso:matrix.org) if you'd like to discuss in more depth, but that's where my heads at right now.

PR #827, opened 2025-03-22
Unify JS files into miso.js

dmjio:
- Combines all 4 js files into one.
- Updates tests to reflect this
- `diff.tests.js` -> `miso.spec.js`


dmjio (commented):
LGTM 👍🏼

PR #828, opened 2025-03-23
Exorcise window from the codebase

dmjio:
- [x] No more hanging things off `window` 
- [x] Attempt to re-export via `module.exports`
- [x] Make `jsaddle` behave
- [x] ~~Conditionally` use ES modules?~~

### DIFF
--- a/jsbits/miso.js
+++ b/jsbits/miso.js
@@ -1,7 +1,7 @@
 /* module export pattern */
 var module = module || {};
 
-module.exports = (function () {
+module['exports'] = (function () {
   /* virtual-dom diffing algorithm, applies patches as detected */
   var diff = function (currentObj, newObj, parent, doc) {
     if (!currentObj && !newObj) return;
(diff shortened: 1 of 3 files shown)


jsbits/miso.js
```diff
@@ -733,71 +731,75 @@ module.exports = (function () {
...
+    'delegate' : delegate,
+    'undelegate' : undelegate,
+    'eventJSON' : eventJSON,
+    'setBodyComponent' : setBodyComponent
   };
 
 })();
+
+/* keep closure compiler from renaming things */
+window['module'] = module;
```
dmjio, line 805: Unfortunately we need to keep a single reference here so that the closure compiler (when using ADVANCED optimizations doesn't rename `module` to something else).

dmjio: But, we are no longer defining functions top-level in terms of window, but rather using a module-export pattern. When not compiling with closure-compiler (or using `node`, `module.exports` is accessible).

PR #829, opened 2025-03-24
Introduce 1ms delay during unmount.

dmjio:
In scenarios (like the `components` example) where deeply nested components undergo unmounting in rapid succession, a delay must be introduced to ensure that unmount actions are processed by the parent before the thread is killed. Otherwise, this can lead to scenarios where a middle child does not have enough time to process both the unmount action from its own child and the relaying of its unmount action to its parent.

Introducing a `threadDelay` will both buy us time and invoke the thread scheduler to ensure other thread stacks have time to evaluate thunks on their heaps before their own destruction (this includes thunks that write to parent sinks).

Tested on both JS and WASM (single threaded environments, and `-threaded` with `jsaddle-warp`)

### DIFF
--- a/src/Miso/Internal.hs
+++ b/src/Miso/Internal.hs
@@ -28,7 +28,7 @@ module Miso.Internal
   ) where
 -----------------------------------------------------------------------------
 import           Control.Exception (throwIO)
-import           Control.Concurrent (ThreadId, killThread)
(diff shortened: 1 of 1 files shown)


dmjio:
A more holistic solution entails sharing `Waiter` (synch. primitives) between parent and child.

PR #830, opened 2025-03-24
var module = module || {}; -> module = {};

dmjio:
By removing `var` we keep `module` named `"module"` top-level and exported globally. It no longer gets GC'd in the WASM context. Also allows us to truly drop `window` usage from miso.js, and not use it in the FFI calls.

### DIFF
--- a/jsbits/miso.js
+++ b/jsbits/miso.js
@@ -1,5 +1,6 @@
 /* module export pattern */
-var module = module || {};
+module = {};
+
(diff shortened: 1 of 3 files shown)


dmjio:
`module = typeof module === 'undefined' ? {} : module;` <- this does the trick to allow the test runner (which uses `node.js`) to work harmoniously with the JS backend and WASM / jsaddle workflows.

PR #836, opened 2025-03-26
Use Double for clientx / clienty

dmjio:
- Use Double for clientx / clienty


dmjio:
https://github.com/dmjio/miso/pull/839

PR #852, opened 2025-04-03
Add production flag for production-grade JS

dmjio:
- Adds `bun build --production` to create `js/miso.prod.js`
- Make `miso-prod-js` derivation.
- Re-add `js/miso.js`, `js/miso.prod.js` "binaries" to `git`
- Add CPP for production flag (not default), for saddle workflows
- Add `build-prod` to package.json
- Put `bun run build-prod` in CI, along with `miso-prod-js`

We have to commit both (prod `js/miso.prod.js` and dev `js/miso.js`) "binaries" to source control now. This is because GitHub users that fetch `miso` from `cabal.project` need the js present in order to build. Adding the production flag is a bonus, but this will work with.


dmjio:
This one is for you @mastarija

mastarija:
> This one is for you @mastarija

Thanks man. I got it twerking ;)

image

dmjio:
Well done

PR #861, opened 2025-04-08
Modernize file reader example, test ergonomics

dmjio:
- Use [miso-from-html](https://github.com/dmjio/miso-from-html) to process claude.ai generated bulma css
- Drop old FFI statements, port to `jsaddle`
- Update cabal file to support WASM and native workflows
- Use `Miso.Lens` for (info field)
- Test `import Miso` ergonomicity
- Use custom css `Style` and bulma cdn for `Href`
- Expose additional `Miso.FFI`, `syncCallback`, etc
- Drop `ghcjs-base`

### DIFF
--- a/examples/file-reader/Main.hs
+++ b/examples/file-reader/Main.hs
@@ -1,106 +1,163 @@
-{-# LANGUAGE CPP #-}
-{-# LANGUAGE DataKinds #-}
-{-# LANGUAGE DeriveGeneric #-}
+-----------------------------------------------------------------------------
+{-# LANGUAGE CPP             #-}
 {-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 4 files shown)


examples/file-reader/Main.hs
```diff
@@ -1,106 +1,163 @@
...
-        setOnLoad reader =<< do
-            asyncCallback $ do
-                r <- getResult reader
-                putMVar mvar r
-        readText reader file
-        SetContent <$> readMVar mvar
-updateModel (SetContent c) = modify $ \m -> m { info = c }
-
--- | View function, with routing
+updateModel ReadFile = scheduleIO $ do
```
dmjio, line 86: This is a little gross, we definitely need a JS quasiquoter here

PR #878, opened 2025-04-16
More fetch fixes

georgefst:
The first one is a basic correctness issue.

The second may look like it doesn't really matter, but it actually saved me from a tricky CORS issue. And AFAICT, there is really no point in adding the header when there's no actual content.


dmjio:
Good catch on the `QueryParam`

PR #882, opened 2025-04-17
Expose more of the fetch API

georgefst:
As discussed just now on Matrix.

For the type class, there's an argument for only exposing it through some internal module rather than from the top-level `Miso` module where it would clutter the haddocks with a pretty niche part of the API. I'll leave that up to @dmjio. Personally I'd seriously consider factoring out the whole fetch API in to a separate library, not tied to Miso, but that suggestion was, not unreasonably, firmly rejected.

### DIFF
--- a/src/Miso/Fetch.hs
+++ b/src/Miso/Fetch.hs
@@ -57,7 +57,7 @@
 ----------------------------------------------------------------------------
 module Miso.Fetch
   ( -- * Class
-    Fetch (fetch)
+    Fetch (..)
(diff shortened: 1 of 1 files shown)


dmjio (approved):

dmjio:
For the record, there are separate fetch libraries that can be used, but I believe `miso` needs to have a featureful fetch / xhr solution internally as well that is part of the `Miso` module. Also, reducing dep. count is a goal too.

PR #888, opened 2025-04-21
Further generalise the fetch API

georgefst:
This has enabled me to work with some fairly big and complex APIs. I'm happy to change some details. Importantly, `fetchJSON` should still work exactly as before, and the internal version, which I've called `fetchFFI`, isn't even exposed.

The only big thing still missing is proper handling of HTTP status codes, but I haven't had a use case for that. Maybe I'll revisit it when the next release of Servant arrives with its multi-verb stuff.


miso.cabal
```diff
@@ -200,6 +200,7 @@ library
     base          < 5,
     bytestring    < 0.13,
     containers    < 0.9,
+    generics-sop  < 0.6,
```
dmjio, line 203: 👀

georgefst: Have switched for `sop-core`. See https://github.com/dmjio/miso/pull/888/commits/f38df9c2b757791f502c5b38b134a0a9fc33ad62.

js/miso.js
```diff
@@ -27,7 +27,7 @@ function fetchJSON(url, method, body, headers, successful, errorful) {
     if (!response.ok) {
       throw new Error(response.statusText);
     }
-    return response.json();
+    return response.text();
```
georgefst, line 30: I think this means that we sometimes UTF-8 encode then decode unnecessarily. We should potentially use `.bytes()` instead.

There's no instance for `FromJSVal ByteString` unfortunately. There is `FromJSVal [Word8]` which is isomorphic, but obviously not the most performant data structure... Could do with some benchmarking. It'd be interesting to get some realistic benchmarks on strings anyway - I'm curious about how much `MisoString` helps when IME one is mostly interacting with libraries that use `Text`.

dmjio: We could try `JSON.parse(response.json())`, in the case of the `MIME` being `JSON`. Then we'd rely on the `FromJSVal Value` instance, since that seems to be faster than parsing. Haskell side it should be more performant.

For `OctetStream` we might want to consider trying `ArrayBuffer` (`.arrayBuffer()`, `.blob()` methods). There's also `.formData()` since I think `servant` supports that as well

georgefst: > Then we'd rely on the `FromJSVal Value` instance, since that seems to be faster than parsing, at least Haskell side it should be more performant.

I guess browsers probably have faster JSON parsers than Aeson, so yeah we could in effect do the `ByteString -> Aeson.Value` step there by using `.json()`. Whether it's worth the cost of doing FFI on a more complex object, I don't know. That part of JSaddle is opaque to me at this point. And anyway, it would have the downside mentioned in the previous comment, i.e. increased complexity.

georgefst: I'd say we find some way to operate on the raw bytes, and save special handling for JSON etc. as a potential future optimisation.

src/Miso/Fetch.hs
```diff
@@ -160,41 +187,94 @@ instance (Fetch api, KnownSymbol name) => Fetch (QueryFlag name :> api) where
...
+      mimeUnrenders :: BSL.ByteString -> NP ([] :.: Either (MediaType, String)) as
+      mimeUnrenders body_ = cpure_NP (Proxy @(UnrenderResponse contentTypes))
+        $ Comp $ unrenderResponse (Proxy @contentTypes) body_
+      tryParsers :: NP ([] :.: Either (MediaType, String)) xs -> Either [String] (Union xs)
+      tryParsers = \case
+        Nil -> Left ["no parsers"]
+        Comp x :* xs -> case partitionEithers x of
+          (err', []) -> bimap (map (\(t, s) -> show t <> ": " <> s) err' <>) S $ tryParsers xs
+          (_, (res : _)) -> Right . inject . I $ res
+class UnrenderResponse (cts :: [Type]) (a :: Type) where
```
dmjio, line 276: Should we use `AllMimeUnrender` w/ `MimeUnrender` from 

https://hackage.haskell.org/package/servant-0.20.2/docs/Servant-API-ContentTypes.html#t:MimeUnrender

georgefst: Yes, good point, actually we can! Just pushed as https://github.com/dmjio/miso/pull/888/commits/59b85a1474e322b4209180d8cac434dbecf79ed9.

I was copying `servant-client` a bit too closely here. I guess they're interesting enough to need their own type class for this, but we're not.

dmjio:
We'll come back to this during native development as we see what that platform can handle.

georgefst:
> @georgefst how should we best proceed w/ this one

Personally, I've abandoned this in favour of following @amesgen's lead from https://github.com/dmjio/miso/pull/904, i.e. I've started using `servant-client-js` instead. It's good enough, although we have discussed forking it in order to iterate more quickly, and in particular to get efficient marshalling of bytes in Wasm.

As for a _non-Servant_ fetch API within Miso, there might still be some stuff above worth salvaging.

PR #889, opened 2025-04-21
Updates JS backend nix instructions

dmjio:
Per feedback, we should ensure we're using miso's pinned nixpkgs for fetching ghcjs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,10 +421,11 @@ $ http-server app.wasmexe
 Using [GHCup](https://www.haskell.org/ghcup/) you should be able to acquire the latest GHC JS-backend compiler.
 
 > [!TIP]
-> For [Nix](https://nixos.org) users it is possible to acquire the latest JS backend via Nix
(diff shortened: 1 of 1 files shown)


dmjio:
cc @akegalj 

```bash
$ cachix use haskell-miso-cachix
$ nix-shell -p pkgs.pkgsCross.ghcjs.haskell.packages.ghc9122.ghc -I nixpkgs=https://github.com/alexfmpe/nixpkgs/archive/b594b289740a2bc917ed9c66fef5d905f389cb96.tar.gz
```

If you're on Darwin it might not be cached, so I will try to push that later today, otherwise a rebuild should succeed.

akegalj:
Thanks, that works like a charm. By the way I used to have `haskell-miso` cache before 
```
https://miso-haskell.cachix.org
miso-haskell.cachix.org-1:6N2DooyFlZOHUfJtAx1Q09H0P5XXYzoxxQYiwn6W1e8=
```
is that one still used or should I drop it in favour of haskell-miso-cachix ?

PR #890, opened 2025-04-21
Add an OSX builder

dmjio:
- This will help us cache stuff for Darwin users
- Push darwin binaries to our cachix cache so new users won't have to build ghcjs9122


dmjio:
Will re-open later, unsure why GitHub runners for `macos-latest` are not available.

PR #898, opened 2025-04-28
Add soup emoji function alias, make its documentation depend on miso

liamzee:
Add soup emoji as an alias for miso. Make its description and miso's description depend on a common chunk.


liamzee:
Documentation issues are extant. I'll have to fix it and repatch it.

PR #900, opened 2025-04-28
Improve Lens haddocks

jhrcek:
What this does: use single block of code for each example instead of multiple separate ">" blocks
Also made few examples simpler/more uniform and added haddock links where appropriate.

Before (left) vs. After
Screenshot from 2025-04-28 17-07-27


dmjio:
Beautiful 👌🏼

PR #905, opened 2025-05-03
Allow isolated, nameless components, introduce 'component_'

dmjio:
This allows us to generate component names on user's behalf when they do not need inter-component communication.

- Add global `IORef` of component IDs, w/ fresh component Id
- If user doesn't give an ID, then generate one from the global ref
- Add `component_` smart constructor
- `setBodyComponent` -> `setComponent`
- Update documentation, removed old comments
- [x] ensure this works with the `delegator`

```haskell
div_ [] [ embed (component_ app) ]
```


src/Miso/FFI/Internal.hs
```diff
@@ -379,8 +380,16 @@ reload = void $ jsg "location" # "reload" $ ([] :: [MisoString])
 setBodyComponent :: MisoString -> JSM ()
 setBodyComponent name = do
   component <- toJSVal name
+  node <- jsg "document" ! "body"
   moduleMiso <- jsg "miso"
-  void $ moduleMiso # "setBodyComponent" $ [component]
+  void $ moduleMiso # "setComponent" $ [node, component]
+-----------------------------------------------------------------------------
+-- | Sets the body with data-component-id
```
jhrcek, line 387: Copy-pasted haddock from setBodyComponent?
```suggestion
-- | Sets @data-component-id@ on the node given by second argument to a value given by the first argument
```

src/Miso/Internal.hs
```diff
@@ -115,6 +115,17 @@ data ComponentState model action
   , componentActions    :: IORef (Seq action)
   }
 -----------------------------------------------------------------------------
+componentIds :: IORef [Int]
+{-# NOINLINE componentIds #-}
+componentIds = unsafePerformIO $ liftIO $ newIORef [ 1 .. ]
```
jhrcek, line 120: Just curious: what's the benefit of having infinite list of IDs as opposed to a single Int value that gets incremented whenever freshComponentId is generated?

src/Miso/Types.hs
```diff
@@ -143,22 +144,32 @@ data SomeComponent
    = forall model action . Eq model
   => SomeComponent (Component model action)
 -----------------------------------------------------------------------------
--- | Used with @component@ to parameterize @App@ by @name@
+-- | A 'Component' wraps an 'App' and can be communicated with via 'componentName'
+-- when using 'notify'. It's state is accessible via 'sample'.
```
jhrcek, line 148:
```suggestion
-- when using 'notify'. Its state is accessible via 'sample'.
```

src/Miso/Types.hs
```diff
@@ -143,22 +144,32 @@ data SomeComponent
...
 data Component model action
   = Component
   { componentKey :: Maybe Key
   , componentName :: MisoString
   , componentApp :: App model action
   }
 -----------------------------------------------------------------------------
--- | Smart constructor for parameterizing @App@ by @name@
+-- | Smart constructor for 'Component' consutruction.
```
jhrcek, line 156:
```suggestion
-- | Smart constructor for 'Component' construction.
```

src/Miso/Types.hs
```diff
@@ -143,22 +144,32 @@ data SomeComponent
...
 -- Needed when calling @embed@ and @embedWith@
 component
   :: MisoString
   -> App model action
   -> Component model action
 component = Component Nothing
 -----------------------------------------------------------------------------
+-- | Smart constructor for 'Component' consutruction.
```
jhrcek, line 164:
```suggestion
-- | Smart constructor for 'Component' construction.
```

src/Miso/Internal.hs
dmjio: Semantically equivalent, performance difference is negligible given the infrequency of usage

dmjio:
A "nameless component" is a `Component` that is isolated from receiving communication with outside world (no other `Component` can send a message to it, but it can send a message to other named components). This is convenient for things like a progress bar that don't need naming and won't necessarily be replaced by other `Component`.

Normally when diffing, we can just replace two nodes if they have different tags, types, keys or `data-component-id`. Or if we're diffing against `null` we know to create or delete a node.

Now, if we allow users to omit the `data-component-id` (ala "nameless components"), and instead we generate it for them, the `data-component-id` no longer becomes a reliable indicator of diffing (since we generate fresh names on each call to `runView` for nested `Component`). This is because we now no longer can tell if a component changed its ID intentionally through `view_` generation, or if it's a placeholder ID that we have generated for the user. Furthermore, a wrench 🔧  is thrown because you can replace a named component with a nameless one (and vice versa).

So here's what I think might be best. Since `Component` shouldn't be high traffic (they shouldn't be moving around the DOM much once they get mounted). We can just tell users that if they plan on replacing `Component` (specifically with other `Component`), they have to rely on `"key"` as a stable name to do so (only in the case of nameless components (i.e. `component_`) that are being replaced by other components -- when using a named component (`component`) we /can/ rely on that name as being stable on diff on it, use it for replacement).

I added a test w/ `Component` replacement diffing that relied on "key" and it has passed (so we keep dom.spec.ts 💯 ). We've added some manual testing to the component example for `Component` replace diffing in the Haskell layer.

But the tldr; is that we're going to have to tell users that if they think it's possible their `Component` will be replaced by another component (when using nameless components), they should give it a "key" (use `embedKeyed_` - same w/ nameless components). Otherwise, if you only expect your `Component` to get created, deleted, replaced by text or node and not replaced by another `Component`, then you can safely ignore the key, when in doubt use `key_`.

This is an edge case, and it most likely won't happen often, when in doubt, name your components (use `component`), if you like to live dangerously, use `component_` but make sure you don't replace it w/ another component in-place, unless you specify a key via `embedKeyed`.

PR #913, opened 2025-05-05
Add dynamic Sub, introduces 'start' and 'stop' for Sub

dmjio:
Allows `Sub` to be created and destroyed dynamically during the lifetime of a `Component` or `App`.

- [x] Introduces `start` and `stop`
- [x] Adds a `DEPRECATION` notice to existing legacy combinators
- [x] Refactors to use `RWS` (R is for the `Component` name)
- [x] Adds `Synchronicity` datatype (reinstates async I/O, explicit `Sync` I/O for drain and `Component` finalisation)
- [x] `withSink` -> `sink`
- [x] Add global `IORef` for dynamic `Sub` ID (top-level `Sub` get autogen'd ID)
- [x] Use `write` vs. `sink` to avoid name shadowing

@brakubraku


src/Miso/Internal.hs
```diff
@@ -387,3 +402,58 @@ renderStyles styles =
...
+        tid <- FFI.forkJSM (sub componentSink)
+        subName <- liftIO freshSubId
+        liftIO $ do
+          atomicModifyIORef' componentSubThreads $ \m ->
+            (M.insert subName tid m, ())
+-----------------------------------------------------------------------------
+-- | 'start'
+--
+-- Starts a named 'Sub' dynamically, during the life of a 'Component'.
+-- The 'Sub' can be stopped by calling 'stop subName' from the 'update' function.
```
jhrcek, line 429: Nit: this is not a link, but code example
```suggestion
-- The 'Sub' can be stopped by calling @stop subName@ from the 'update' function.
```

src/Miso/Internal.hs
```diff
@@ -240,7 +255,7 @@ unmount
 unmount mountCallback app@App {..} cs@ComponentState {..} = do
   undelegator componentMount componentVTree events (logLevel `elem` [DebugEvents, DebugAll])
   freeFunction mountCallback
-  liftIO (mapM_ killThread componentSubThreads)
+  liftIO $ mapM_ killThread =<< M.elems <$> liftIO (readIORef componentSubThreads)
```
jhrcek, line 258: No need to extract list of elements from the map, since Map itself is traversable
```suggestion
  liftIO $ mapM_ killThread =<< readIORef componentSubThreads
```

PR #917, opened 2025-05-06
Use IORef Int instead of IORef [Int] for global vars

dmjio:
Use `Int` accumulator instead of an infinite lazy list


dmjio:
@jhrcek

PR #918, opened 2025-05-06
sink -> withSink

dmjio:
Use more idiomatic `with` prefix when dealing when continuations


jhrcek (approved):
I like this more than previous version :+1:

dmjio:
yea, it ultimately makes more sense, `sink` was too cute, and ultimately non-intuitive

PR #919, opened 2025-05-06
Remove deprecated and add newly introduced SVG attributes

jhrcek:
Followup on https://github.com/dmjio/miso/pull/914

I went through https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute and removed all the attributes for which the docs are no longer available or are explicitly marked as deprecated.


dmjio:
Hero 🏆

Issue #746, opened 2024-10-04
The best way to handle style sheets in Miso application or component library


functora:
I would like to discuss the best way to handle style sheets in Miso application or component library. I think styles abstractions (whatever it is) should be:

- Composable with library/application Haskell code.
- Composable with each other.
- More or less typed (optional).

At the moment I see couple of options, but they all are kinda bad:

1) Using the standard external CSS file. The problems are:
- I don't see the right way to export it as a part of the library (these are no `css-sources` in cabal, unlike `js-sources`)
- CSS class names are just Text which should be the same across Haskell (`class_`) and CSS codebase. Hard to maintain and refactor. Prone to errors.
2) Using inline `<style>` tags next to every small component. I think kinda bad idea, and I'm not sure how to scope styles locally to component this way. Not sure it is possible or easy (not an expert in CSS). 
3) Using `style_` attribute functions and different variants of compositions of it. It's a bit better, but:
- Can not define styles for pseudo-elements (::hover, ::before, ::after etc)
- Can not define advanced CSS animations like `@keyframes my-animation-name`
4) Doing some smart CSS code generation from some Haskell style sheet source (.hs files), exporting class names (functions, utilities) used in Miso lib/app codebase, but having additional Setup step for the app/lib which doing corresponding CSS file code generation somehow.
- Kinda sounds hard to do (maybe not, idk).

If I did miss some documentation or cool advanced CSS examples in Miso, please let me know. Best regards!

dmjio:
This is an important topic. There's a lot of different schools of thought on how to create and manage CSS in applications (CSS-in-JS, CSS preprocessor, SMACSS, BEM, tailwind). In an attempt to please everybody we did not make any hard decisions.

With that said, for preprocessors there is a project called `clay` that could be integrated with miso, providing additional type safety. Also I believe there are attempts to wrap tailwind classes with miso's class attribute (otherwise wouldn't be difficult to do this on your own).

dmjio:
@functora http://fvisser.nl/clay/ from @sebastiaanvisser

dhess:
> Also I believe there are attempts to wrap tailwind classes with miso's class attribute (otherwise wouldn't be difficult to do this on your own).

Hi @dmjio, thanks so much for this project. Do you have any references for those Tailwind+Miso integration attempts?

dmjio:
@functora, 

I have come up with a partial solution for CSS in this PR https://github.com/dmjio/miso/pull/859

[...]

PR #920, opened 2025-05-06
Adds Miso.Style

dmjio:
- [ ] Add additional combinators / strong types
- [x] Update examples to use new `style_` combinator
- [x] Adds ISO colors via `Miso.Style.Color`
- [x] Preserve existing `style_` HTML combinator (this is why `Miso.Style` must be used qualified)

Addresses https://github.com/dmjio/miso/issues/746

@jhrcek

### DIFF
--- a/src/Miso/Property.hs
+++ b/src/Miso/Property.hs
@@ -1,0 +1,54 @@
+-----------------------------------------------------------------------------
+{-# LANGUAGE OverloadedStrings #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Miso.Property
+-- Copyright   :  (C) 2016-2025 David M. Johnson
+-- License     :  BSD3-style (see the file LICENSE)
+-- Maintainer  :  David M. Johnson <code@dmj.io>
+-- Stability   :  experimental
+-- Portability :  non-portable
+--
+-- Construct custom properties on DOM elements
+--
+-- > div_ [ prop "id" "foo" ] [ ]
+--
+----------------------------------------------------------------------------
+module Miso.Property
+  ( -- *** Smart constructors
+    textProp
+  , stringProp
+  , boolProp
+  , intProp
+  , integerProp
+  , doubleProp
+  , prop
+  ) where
+-----------------------------------------------------------------------------
+import           Miso.Html.Types
+import           Miso.String (MisoString)
+-----------------------------------------------------------------------------
+-- | Set field to `Bool` value
+boolProp :: MisoString -> Bool -> Attribute action
+boolProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `String` value
+stringProp ::  MisoString -> String -> Attribute action
+stringProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `Text` value
+textProp ::  MisoString -> MisoString -> Attribute action
+textProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `Int` value
+intProp ::  MisoString -> Int -> Attribute action
+intProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `Integer` value
+integerProp ::  MisoString -> Int -> Attribute action
+integerProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `Double` value
+doubleProp ::  MisoString -> Double -> Attribute action
+doubleProp = prop
+-----------------------------------------------------------------------------
--- a/src/Miso/Property.hs
+++ b/src/Miso/Property.hs
@@ -25,8 +25,15 @@ module Miso.Property
   , prop
   ) where
 -----------------------------------------------------------------------------
-import           Miso.Html.Types
+import           Data.Aeson (ToJSON(..))
+-----------------------------------------------------------------------------
+import           Miso.Types
 import           Miso.String (MisoString)
+-----------------------------------------------------------------------------
+-- | @prop k v@ is an attribute that will set the attribute @k@ of the DOM 
+-- node associated with the vnode to @v@.
+prop :: ToJSON a => MisoString -> a -> Attribute action
+prop k v = Property k (toJSON v)
 -----------------------------------------------------------------------------
 -- | Set field to `Bool` value
 boolProp :: MisoString -> Bool -> Attribute action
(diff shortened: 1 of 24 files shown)


dmjio:
@functora

src/Miso/Style.hs
```diff
@@ -0,0 +1,1524 @@
...
+--
+-- > view m = div_ [ styleInline_ "background-color:red;color:blue;" ] [ "foo" ]
+--
+-- https://developer.mozilla.org/en-US/docs/Web/CSS
+styleInline_ ::  MisoString -> Attribute action
+styleInline_ = textProp "style"
+-----------------------------------------------------------------------------
+-- | 'renderStyleSheet'
+--
+-- Renders a 'StyleSheet' to a 'MisoString'
```
jhrcek, line 264: Suggestion: I think adding link with the function name in the haddocks of that function only adds noise to the haddock (nobody is going to every click on the link to function X to navigate to function X if that function is just below it). I would suggest removing these. If you search for "-- | '" it's 155x in this PR :sweat_smile: 
```suggestion
-- | Renders a 'StyleSheet' to a 'MisoString'
```

src/Miso/Style.hs
```diff
@@ -0,0 +1,1524 @@
...
+-- 'flexDirection'
+--
+-- > styles_ [ flexDirection .= "value" ]
+--
+flexDirection :: MisoString -> Style
+flexDirection x = "flex-direction" === x
+-----------------------------------------------------------------------------
+-- 'flexFlow'
+--
+-- > styles_ [ flexFlow .= "value" ]
```
jhrcek, line 778: Is this `.=` operator in haddock correct? It looks like something from aeson or lens..

dmjio: No, good catch. Let me find and replace w/ VSCode

jhrcek: Also all the examples use `styles_` (plural) which doesn't exist.

dmjio: doh

dmjio: fixed

src/Miso/Property.hs
```diff
@@ -0,0 +1,61 @@
...
+-- | Set field to `Text` value
+textProp ::  MisoString -> MisoString -> Attribute action
+textProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `Int` value
+intProp ::  MisoString -> Int -> Attribute action
+intProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `Integer` value
+integerProp ::  MisoString -> Int -> Attribute action
```
jhrcek, line 55:
```suggestion
integerProp ::  MisoString -> Integer -> Attribute action
```

src/Miso/Property.hs
```diff
@@ -0,0 +1,61 @@
...
+integerProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `Double` value
+doubleProp ::  MisoString -> Double -> Attribute action
+doubleProp = prop
+-----------------------------------------------------------------------------
```
jhrcek, line 61:
```suggestion
-----------------------------------------------------------------------------

```

src/Miso/Property.hs
```diff
@@ -0,0 +1,61 @@
...
+-- | Set field to `Bool` value
+boolProp :: MisoString -> Bool -> Attribute action
+boolProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `String` value
+stringProp ::  MisoString -> String -> Attribute action
+stringProp = prop
+-----------------------------------------------------------------------------
+-- | Set field to `Text` value
+textProp ::  MisoString -> MisoString -> Attribute action
```
jhrcek, line 47:
```suggestion
textProp ::  MisoString -> Text -> Attribute action
```

src/Miso/Style.hs
```diff
@@ -0,0 +1,1357 @@
...
+-----------------------------------------------------------------------------
+import           Prelude hiding (filter)
+-----------------------------------------------------------------------------
+-- | Smart constructor for Attributes. This function is helpful when 
+-- constructing 'Style'.
+--
+-- Example shown below.
+--
+-- @
+-- div_ [ style_  [ "background" =: "red" ] []
```
jhrcek, line 221:
```suggestion
-- div_ [ style_  [ "background" =: "red" ] ] []
```

src/Miso/Style.hs
```diff
@@ -0,0 +1,1357 @@
...
+-- | 'Styles'
+--
+-- Type for a @Map@ of CSS 'Style'. Used with @StyleSheet@.
+--
+type Styles = Map MisoString MisoString
+-----------------------------------------------------------------------------
+-- | 'StyleSheet'
+--
+-- Type for a CSS style on native
+--
```
jhrcek, line 243: It's Map from MisoString to Styles, which itself is Map from MisoString to Miso String.
Maybe add a sentence about what the keys / values at each level are? The same for sheet_ below.

src/Miso/Style.hs
```diff
@@ -0,0 +1,1357 @@
...
+-- > view m = div_ [ styleInline_ "background-color:red;color:blue;" ] [ "foo" ]
+--
+-- https://developer.mozilla.org/en-US/docs/Web/CSS
+styleInline_ ::  MisoString -> Attribute action
+styleInline_ = textProp "style"
+-----------------------------------------------------------------------------
+-- | 'renderStyleSheet'
+--
+-- Renders a 'StyleSheet' to a 'MisoString'
```
jhrcek, line 272: This comment belongs to the function below?

src/Miso/Property.hs
dmjio: Let's keep it `Int`, if you want to rename it then do `intProp`

dmjio: but honestly, probably best to just leave it for now, since its a big breaking change

jhrcek: But then the function name and haddock above it are lying.
What's the purpose of having both intProp and integerProp with identical type signature?

src/Miso/Property.hs
jhrcek: in that case just update the Haddock not to mention `Text` :slightly_smiling_face:

dmjio: MisoString becomes Text on GHC and stays JSString on WASM

dmjio:
We'll come back to this, we might take a full lexer / parser approach to CSS. Have to move back onto docs and native, we'll keep fleshing it out piecemeal on-demand.

PR #921, opened 2025-05-07
Use consistent naming for nodeHtml/Svg/Mathml, avoid flip

jhrcek:


src/Miso/Html/Element.hs
```diff
@@ -130,7 +130,7 @@ import           Miso.String (MisoString)
 -----------------------------------------------------------------------------
 -- | Used to construct @Node@ in @View@
 nodeHtml :: MisoString -> [Attribute action] -> [View action] -> View action
-nodeHtml = flip (node HTML) Nothing
```
jhrcek, line 133: I'm not sure whether flip adds any overhead here, but I think it's better to not having to guess, (since these node combinators are used all over the place) and simply avoid it.

PR #922, opened 2025-05-07
Remove deprecated and add newly introduced HTML elements

jhrcek:


src/Miso/Html/Element.hs
```diff
@@ -10,119 +10,138 @@
...
+    -- ** Embedded content
+    , embed_
+    , fencedframe_
     , iframe_
+    , object_
+    , picture_
+    , source_
+    -- ** Scripting
     , canvas_
-    , math_
```
jhrcek, line 53: Highlight: math_ node should live in MATHML namespace, so I moved this helper to Mathml.Element module and reused it in Mathml example.
I think it would make sense to add all supported mathml elements to that module (https://developer.mozilla.org/en-US/docs/Web/MathML/Reference/Element) but I'll do that in separate PR to keep this PR focused.

src/Miso/Html/Element.hs
```diff
@@ -10,119 +10,138 @@
 -- Portability :  non-portable
 ----------------------------------------------------------------------------
 module Miso.Html.Element
-  ( -- *** Smart constructors
+  ( -- ** Smart constructors
       nodeHtml
     , nodeHtmlKeyed
-    -- *** Combinators
-    , doctype_
+    -- ** Document metadata
```
jhrcek, line 16: I grouped the elements in the export list using the same categories that MDN reference uses

dmjio:
Awesome

PR #925, opened 2025-05-09
Use markdown links for more readable haddocks, add @since annotations

jhrcek:


src/Miso/Svg.hs
```diff
@@ -33,7 +33,7 @@ module Miso.Svg
    , module Miso.Svg.Event
    ) where
 -----------------------------------------------------------------------------
-import Miso.Svg.Property hiding (filter_, path_, mask_, clipPath_, cursor_, style_)
```
jhrcek, line 36: Just fixing warnings after style_ was moved in https://github.com/dmjio/miso/pull/920

PR #926, opened 2025-05-09
Component refactor

dmjio:
This should crystallize / simplify the component feature.

- `App` is now parameterized by `name :: Symbol`, this allows `notify` and `sample` to be type safe during message passing and reading other component state. The `name` parameter is optional unless using the component feature (which means using `notify` / `sample`)

- `Component` is now used purely as an existential wrapper for `App`, when used w/ `notify_` / `sample_` its to indicate that the component is dynamically generated. The dynamically generated component can be communicated with now via its name, which is accessible via `ask` (per the `RWS` refactor).

- Introduces `notify_`, `sample_`,

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ foreign export javascript "hs_start" main :: IO ()
 #endif
 ----------------------------------------------------------------------------
 -- | `defaultApp` takes as arguments the initial model, update function, view function
-app :: App Model Action
+app :: App name Model Action
 app = defaultApp emptyModel updateModel viewModel
 ----------------------------------------------------------------------------
 -- | Empty application state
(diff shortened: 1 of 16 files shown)


dmjio:
@brakubraku

dmjio:
```haskell
notify :: App name model action -> action -> JSM ()
notify_ :: MisoString -> action -> JSM () -- unsafe but allow dynamic component message passing

sample :: App name model action -> JSM model
sample_ :: MisoString -> JSM model -- unsafe but allows dynamic component sampling

component :: App name model action -> View a
component_ :: SomeApp -> View a -- used for dynamic component generation
```

PR #929, opened 2025-05-10
More haddock cleanups

jhrcek:
- don't use removed and deprecated stuff in examples (scheduleIO, Transition monad)
- avoid haddocks that just repeat the function name (they just add noise, take up space and don't provide any informative value)


src/Miso/Style.hs
```diff
@@ -267,34 +279,14 @@ style_ = MT.Styles . M.fromList
...
---   ]
--- @
---
+-- | Renders a 'Styles' to a 'MisoString'
 renderStyles :: Styles -> MisoString
 renderStyles m = MS.unlines
   [ mconcat [ spaced, k, ":", v, ";" ]
   | (k,v) <- M.toList m
-  , let spaced = "  "
+  , let spaced = " "
```
jhrcek, line 287: Probably the only non-haddock change - let me know if using 2 spaces was intentional (or is it for aesthetic reasons? Seems unnecessary to increase stylesheet size)

src/Miso/Effect.hs
```diff
@@ -196,15 +197,13 @@ for actions = withSink $ \sink -> actions >>= flip for_ sink
 withSink :: (Sink action -> JSM ()) -> Effect model action
 withSink f = tell [ f ]
 -----------------------------------------------------------------------------
--- | A synonym for @tell@, specialized to @Effect@
```
jhrcek, line 199: This comment seems to have been misplaced, as the types of tell and issue don't match. I moved tit to withSink which looks more like a synonym to tell.

src/Miso/Style.hs
```diff
@@ -267,34 +279,14 @@ style_ = MT.Styles . M.fromList
 styleInline_ ::  MisoString -> Attribute action
 styleInline_ = textProp "style"
 -----------------------------------------------------------------------------
--- | 'renderStyleSheet'
---
--- Renders a 'StyleSheet' to a 'MisoString'
---
--- @
--- testSheet :: StyleSheet
```
jhrcek, line 275: This example seems better placed above StyleSheet definition

dmjio (approved):

PR #933, opened 2025-05-11
Fix haddock issues in Miso.Lens

jhrcek:


src/Miso/Lens.hs
```diff
@@ -324,15 +324,22 @@ infixr 4 <>~
...
+--    ...
 -- @
--- newtype List = List { _values :: [Int] }
 --
--- values :: Lens List [Int]
--- values = lens _values $ \\l vs -> l { _values = vs }
+-- will store the result in a variable, while
 --
--- addElement :: List -> List
--- addElement list = list & values <>~ [2]
```
jhrcek, line 334: This was copy pasted example from <>~.
It's not as easy to come up with a self-contained example for `<~`, so I just copied its explanation from the lens package.

PR #934, opened 2025-05-11
Add element and attribute helpers for MathML

jhrcek:


examples/mathml/Main.hs
```diff
@@ -22,11 +23,38 @@ updateModel _ = pure ()
 viewModel :: Model -> View ()
 viewModel _ =
     math_
-        []
-        [ nodeMathml
-            "msup"
-            []
-            [ nodeMathml "mi" [] [text "x"]
-            , nodeMathml "mn" [] [text "2"]
+        [display_ "block"]
```
jhrcek, line 26: Changed this to something slightly more complex (quadratic formula).
Works fine for me in chrome and firefox
Screenshot from 2025-05-11 15-32-49

src/Miso.hs
```diff
@@ -48,8 +48,6 @@ module Miso
     -- * Html
   , module Miso.Html
   , module Miso.Render
-    -- * Mathml
-  , module Miso.Mathml
```
jhrcek, line 52: I think it doesn't make much sense to export Mathml modules from the top-level Miso module.
1. it will be likely relatively rarely used
2. there's bunch of attributes that clashes with html/svg attributes - so if users do want to math stuff, they should import it separately.

src/Miso/Mathml/Property.hs
```diff
@@ -0,0 +1,229 @@
+-----------------------------------------------------------------------------
+{-# LANGUAGE OverloadedStrings #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Miso.Html.Property
```
dmjio, line 5:
```suggestion
-- Module      :  Miso.Html.Property
-- Module      :  Miso.Mathml.Property
```

src/Miso/Mathml/Element.hs
```diff
@@ -12,17 +12,217 @@
...
+  , munder_
+  , munderover_
+  , semantics_
   ) where
 -----------------------------------------------------------------------------
 import           Miso.Html.Types
 import           Miso.String (MisoString)
 -----------------------------------------------------------------------------
--- | Used to construct @Node@ in @View@
+-- | Low-level used to construct @Node@ in @View@.
```
jhrcek, line 50:
```suggestion
-- | Low-level function used to construct @Node@ in @View@.
```

examples/mathml/Main.hs
dmjio: hell yes

PR #935, opened 2025-05-11
Add Space Invaders demo

dmjio:
Dear @juliendehos,

Thanks for the `miso-invaders` example. I'd like to add it to the README.md.

P.S. is there any way we can change the audio track :)


juliendehos:
@dmjio 
Thanks for adding this example to the README.md.
I've updated the code with the latest version of Miso (the Canvas API and WASM backend are awesome).
I've also changed the audio track but if you have any better ideas, please let me know.

juliendehos:
Thanks for your comments.
I tried `Miso.Lens`, but it wasn't very convenient to use with the `Linear` package, as the latter uses `Lens`.
I'm rewriting the audio portion of the example; I'll join the Matrix chat to discuss it.

Issue #915, opened 2025-05-05
Declarative 2D / 3D canvas support


dmjio:
It would be nice to have something like thres or threlte exposed so we can have declarative canvas animations

- https://tresjs.org/
- https://threlte.xyz/

- [x] 2D Canvas
- [ ] WebGL (OpenGL ES)

dmjio:
https://dannymcgee.dev/posts/declarative-canvas-web

fivemeyestore:
This is a well-structured codebase. I appreciate the clean separation of concerns.

godofcodeeyes:
This is a well-structured codebase. I appreciate the clean separation of concerns.

dmjio:
https://hackage.haskell.org/package/blank-canvas

dmjio:
Canvas is in, going to put three.js into a separate. Request to re-open if you want WebGL `Miso.WebGL`

PR #943, opened 2025-05-15
Rewrite three.js example to use jsaddle, WASM-backend.

dmjio:
- [x] Adds scripts for building three.js w/ WASM backend
- [x] Port three.js FFI to jsaddle
- [x] Drop cabal js-only option in miso-examples.cabal


dmjio:
Related #915

PR #945, opened 2025-05-18
Abstract over canvas dimension, interpretation.

dmjio:
This PR introduces a callback to the `canvas` function that can be used to abstract over 2d and 3d canvas draws.

This is the first step towards a declarative API for WebGL / three.js.


dmjio:
#915

PR #946, opened 2025-05-19
Fix type signatures in startSub, stopSub.

brakubraku:


dmjio:
nice

dmjio:
the components example didn't catch this because `loggerSub` was kept generic

PR #948, opened 2025-05-20
Add key_, drop componentWith* variants.

dmjio:
`Key` has always existed as an explicit identifier internally, this change makes` Key` no different than any other 'Attribute'. User's can now set a `Key` inside of an `Attribute` list using 'key_'. This change is purely cosmetic for the Haskell interface, no TypeScript changes required. ~This preserves existing `trKeyed_`, `liKeyed_` combinators as well~

```haskell
-- example
keyedDiv = div_ [ key_ 1 ] []
```

- [x] Introduce `key_`
- [x] Drop `Key` parameter internally
- [x] Set `"key"` inside of `'setAttrs'` internally (in `runView`)
- [x] Use `textProp` in `parseView`
- [x] `ToJSON` on `Key`
- [x] Drop `with*` variants of `component`.
- [x] Drop `trKeyed`, `liKeyed` variants

### DIFF
--- a/src/Miso/Internal.hs
+++ b/src/Miso/Internal.hs
@@ -37,7 +37,6 @@ import           Control.Exception (throwIO)
 import           Control.Monad (forM, forM_, when, void)
 import           Control.Monad.Reader (ask)
 import           Control.Monad.IO.Class
-import qualified Data.Aeson as A
 import           Data.Foldable (toList)
 import           Data.IORef (IORef, newIORef, atomicModifyIORef', readIORef, atomicWriteIORef, modifyIORef')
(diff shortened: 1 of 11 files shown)


src/Miso/Internal.hs
```diff
@@ -391,6 +390,9 @@ setAttrs
   -> JSM ()
 setAttrs vnode attrs snk logLevel events =
   forM_ attrs $ \case
+    Property "key" v -> do
+      value <- toJSVal v
+      FFI.set "key" value vnode
```
dmjio, line 395: ☝🏼

PR #952, opened 2025-05-22
Generalize SubName to Ord key => key.

dmjio:
- [x] Update `startSub`, `stopSub`
- [x] Update component example, `LoggerSub`
- [x] Drop `SubName`

### DIFF
--- a/src/Miso/Internal.hs
+++ b/src/Miso/Internal.hs
@@ -463,7 +463,13 @@ renderStyles styles =
 -- The 'Sub' can be stopped by calling @stop subName@ from the 'update' function.
 -- All 'Sub' started will be stopped if a 'Component' is unmounted.
 --
-startSub :: Ord key => key -> Sub action -> Effect model action
+-- @
+-- data SubType = LoggerSub | TimerSub
+--
+-- update Action =
+--   startSub LoggerSub $ \sink -> forever (threadDelay (secs 1) >> consoleLog "test")
+-- @
+startSub :: Ord sub => sub -> Sub action -> Effect model action
 startSub subName sub = do
   compName <- ask
   io_
@@ -489,7 +495,13 @@ startSub subName sub = do
 -- | Stops a named 'Sub' dynamically, during the life of a 'Component'.
 -- All 'Sub' started will be stopped automatically if a 'Component' is unmounted.
 --
-stopSub :: Ord key => key -> Effect model action
+-- @
+-- data SubType = LoggerSub | TimerSub
+--
+-- update Action = do
+--   stopSub LoggerSub
+-- @
+stopSub :: Ord sub => sub -> Effect model action
 stopSub key = do
   compName <- ask
   io_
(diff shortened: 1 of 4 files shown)


src/Miso/Internal.hs
```diff
@@ -463,7 +463,7 @@ renderStyles styles =
 -- The 'Sub' can be stopped by calling @stop subName@ from the 'update' function.
 -- All 'Sub' started will be stopped if a 'Component' is unmounted.
 --
-startSub :: SubName -> Sub action -> Effect model action
```
jhrcek, line 466: >  The 'Sub' can be stopped by calling @stop subName@ from the 'update' function

Could you please also update mentions of subName in haddocks?

src/Miso/Internal.hs
```diff
@@ -122,10 +122,10 @@ data Hydrate
   deriving (Show, Eq)
 -----------------------------------------------------------------------------
 -- | @Component@ state, data associated with the lifetime of a @Component@
-data ComponentState model action
+data ComponentState key model action
```
jhrcek, line 125: I think the key type param should be explained at the minimum in the ComponentState haddocks.
There's a high risk of people confusing this particular key param with the "Key" used for speeding up vdom diffing.

dmjio: I will in a latter pass, but also this isn't a part of the public API

src/Miso/Internal.hs
dmjio: `subName` can now mean `Ord subName => subName`, so it's technically still valid

src/Miso/Internal.hs
```diff
@@ -463,7 +463,13 @@ renderStyles styles =
 -- The 'Sub' can be stopped by calling @stop subName@ from the 'update' function.
 -- All 'Sub' started will be stopped if a 'Component' is unmounted.
 --
-startSub :: SubName -> Sub action -> Effect model action
+-- @
+-- data SubType = LoggerSub | TimerSub
+--
+-- update Action =
+--   startSub LoggerSub $ \sink -> forever (threadDelay (secs 1) >> consoleLog "test")
```
jhrcek, line 470:
```suggestion
-- data SubType = LoggerSub | TimerSub deriving (Eq, Ord)
--
-- update Action =
--   startSub LoggerSub $ \\sink -> forever (threadDelay (secs 1) >> consoleLog "test")
```

src/Miso/Internal.hs
```diff
@@ -489,17 +495,23 @@ startSub subName sub = do
 -- | Stops a named 'Sub' dynamically, during the life of a 'Component'.
 -- All 'Sub' started will be stopped automatically if a 'Component' is unmounted.
 --
-stopSub :: SubName -> Effect model action
-stopSub subName = do
+-- @
+-- data SubType = LoggerSub | TimerSub
```
jhrcek, line 499:
```suggestion
-- data SubType = LoggerSub | TimerSub deriving (Eq, Ord)
```

src/Miso/Internal.hs
dmjio: used `subKey`

PR #980, opened 2025-06-11
Update README related to `post-link.mjs`

rinn7e:
Problem: Currently, `post-link.mjs` won't work properly, if the node version is not up-to-date.

Solution: Add a note about nodejs version in `post-link.mjs` section.


dmjio (changes requested):
Looks good, can we make it a `TIP` so it pops out a bit at the user

```markdown
> [!TIP]
> We recommend using an up-to-date `node` version (currently tested with `v24.2.0`) to ensure `post-link.mjs` works properly.
```

README.md
```diff
@@ -330,6 +330,9 @@ But, we're not done yet. In order to view this in the browser there are still a
 
 To start, we recommend creating an `app.wasmexe` folder to store the additional artifacts required.
 
+> [!TIP]
+> We recommend using an up-to-date `node` version (currently tested with `v24.2.0`) to ensure `post-link.mjs` works properly.
+
```
rinn7e, line 335: @dmjio I've pushed the fix according to your suggestion.

dmjio: LGTM

PR #990, opened 2025-06-20
Fix

brakubraku:


ts/miso/dom.ts
brakubraku, line 1: I may have sniffed out the offender...

ts/miso/dom.ts
```diff
@@ -134,7 +134,7 @@ function diffProps(cProps: Props, nProps: Props, node: Element, isSvg: boolean,
     if (newProp === undefined) {
       /* current key is not in node, remove it from DOM, if SVG, remove attribute */
       if (isSvg || !(c in node)) {
-        context['removeAttribute'](node, cProps[c]);
```
dmjio, line 137: That's definitely a bug 🐛

dmjio: Well done, great find @brakubraku 🏆 will follow-up w/ a test.

dmjio: For additional context, this bug attempted to remove the property value instead of its name. Unfortunately the tests didn't catch this because the test corpus was using the same string for both key and value. That has been solved in https://github.com/dmjio/miso/commit/cdda1144d979ccdb3872db70e5e9c48069cae7b4

PR #996, opened 2025-06-23
Adds a `flake.nix` for `miso`.

dmjio:
### Miso Flakes

Now that native supports `flake.nix`, we can upstream to `miso`.

- [x] Adds `flake.nix` to support a `flake` workflow
- [x] Include `overlays.nix`.
- [x] Include el cache
- [x] Consider re-exporting `ghc-wasm-meta` flake
- [x] Consider exposing GHCJS-9122 / GHCJS-86 as well (names shouldn't collide)
- [x] Consume `flake.lock` in `source.nix`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -59,13 +59,30 @@
 
         # Miso's packages
         packages = rec {
-          default = miso;
-          inherit (pkgs)
-            miso
-            miso-examples
-            miso-ghc
-            miso-from-html
+          # Default package is vanilla GHC 9.12.2 miso
+          default = miso-ghc-9122;
+
+          # GHCJS miso, miso-examples
+          miso-ghcjs-9122 =
+            pkgs.pkgsCross.ghcjs.haskell.packages.ghc9122.miso;
+          miso-examples-ghcjs-9122 =
+            pkgs.pkgsCross.ghcjs.haskell.packages.ghc9122.miso-examples;
+
+          # GHC
+          miso-ghc-9122 =
+            pkgs.haskell.packages.ghc9122.miso;
+
+          miso-examples-ghc-9122 =
+            pkgs.haskell.packages.ghc9122.miso;
+
+          # Util
+          inherit (pkgs.haskell.packages.ghc9122)
+            miso-from-html;
+
+          # Misc.
+          inherit (pkgs.haskell.packages.ghc9122)
             more-examples;
+
         };
 
         # Miso's dev shells
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,53 @@
+{
+  description = "A flake for miso";
+
+  inputs = {
+
+    nixpkgs.url =
+      "https://github.com/alexfmpe/nixpkgs/archive/b594b289740a2bc917ed9c66fef5d905f389cb96.tar.gz";
+
+    flake-utils.url =
+      "github:numtide/flake-utils";
+
+  };
+  outputs = { self, nixpkgs, flake-utils }:
+
+   flake-utils.lib.eachSystem
+      [
+        "x86_64-linux"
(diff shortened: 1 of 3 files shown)


dmjio:
Sure feel free, sounds great to me, ideally we could have the WASM flake, the cached GHCJS (and soon modified one for `miso-native`), and GHC all in the shell via `nix develop`, and buildable as well.. I had issues re-exporting the WASM flake from `ghc-wasm-meta`. 

Whatever you think flake users expect in their workflow most I'd like to accommodate @srid

srid:
Great.

#1009 can be merged right away. 

I'll take a look at the haskell-flake part later.

dmjio:
We need to get the flake working first, which means putting the WASM backend and GHCJS compilers in `$PATH`

srid:
You can do something like this. Then `nix develop` puts you in ghc shell, and `nix develop .#ghcjs` should put you in the ghcjs one. 

```diff
diff --git a/flake.nix b/flake.nix
index 6b94457..4629979 100644
--- a/flake.nix
+++ b/flake.nix
@@ -38,11 +38,9 @@
 
         packages.default = pkgs.haskell.packages.ghc9122.miso;
 
-        devShells.default = pkgs.mkShell {
-          name = "The miso ghc9122 shell";
-          buildInputs = with pkgs; [
-            pkgs.pkgsCross.ghcjs.haskell.packages.ghc9122.miso
-          ];
+        devShells = pkgs.mkShell {
+          default = pkgs.haskell.packages.ghc9122.miso.env;
+          ghcjs = pkgs.pkgsCross.ghcjs.haskell.packages.ghc9122.miso.env;
         };
       };
     };
```

But the later doesn't work due to:

```
       error: Package ‘ghcjs-base-0.8.0.2’ in /nix/store/7bqmq5iqvkbpm5qbq62hm072wzm7hj61-source/pkgs/development/haskell-modules/hackage-packages.nix:122356 is not available on the requested hostPlatform:
         hostPlatform.config = "x86_64-unknown-linux-gnu"
         package.meta.platforms = [
           "javascript-ghcjs"
         ]
         package.meta.badPlatforms = [ ]
       , refusing to evaluate.
```


How does the legacy `shell.nix` dev workflow work? Is the idea to port that over to flakes? Or write something anew?

dmjio:
@srid If we do `ghcjs = pkgs.pkgsCross.ghcjs.haskell.packages.ghc9122.ghc` that might be good enough to get us a shell w/ the latest GHCJS9122 backend (we can probably omit `ghcjs86` for end users, but all of our code is backwards compatible to `ghcjs86`).

@terrorjack's WASM backend [supports GHCi now](https://www.tweag.io/blog/2024-11-21-ghc-wasm-th-ghci/), so we should be able to re-export from `ghc-wasm-meta`'s flake's shell, if that's supported.

dmjio:
cc @amesgen

flake.nix
```diff
@@ -0,0 +1,169 @@
...
+                 inputs.ghc-wasm-meta.packages.${system}.all_9_12
+                 pkgs.gnumake
+                 pkgs.http-server
+                 pkgs.cabal-install
+               ];
+              shellHook = ''
+                function build () {
+                   wasm32-wasi-cabal build $1 \
+                     --with-compiler=wasm32-wasi-ghc \
+                     --with-hc-pkg=wasm32-wasi-ghc-pkg
```
amesgen, line 132: The `--with-compiler=wasm32-wasi-ghc --with-hc-pkg=wasm32-wasi-ghc-pkg` shouldn't actually be necessary, or did you find that it was?

dmjio: You’re right it’s not necessary we can drop this.

PR #1001, opened 2025-07-01
Initial commit of intergration with `three.hs`

dmjio:
Adds [three.hs](github.com/three-hs/three.hs/) integration


dmjio:
`miso` now exports `canvas`, which is integrated with https://github.com/three-hs/ per work by @juliendehos . Future three.js work will be done there, and has been factored out of `miso` cleanly.

Issue #1000, opened 2025-06-30
Components as values makes modularity with parent/child components awkward


kaol:
I've done some code using components with using master version and I would have some feedback on the way how components are defined, with regard to using subcomponents. The way how the functions that target components, like `notify`, take a value of type `Component Model Action` feels awkward if the components are defined in different modules.

How should child components access parents in this model across modules? It's natural to design a parent component in a module that imports subcomponents that are defined in other modules, but if a submodule needs to send an Action to the parent, it needs the concrete value for it and if that's in another module this would introduce an import cycle.

This situation would arise if you'd try to move `counterComponent4` to a different module from `examples/components/Main.hs`. Doing that would lose easy access to `counterComponent3`.

My instinct in a situation like this would be to move the parent's Action to a different module that both the parent and the child would import separately, breaking the loop. At least for `notifiy`, only the component name and parent's `Action` is needed, all the rest are accessed from the hidden global state. Using the component value for this operation includes the view render function and that pulls a lot of other dependencies.

One idea about how to use these functions that I can think of is to use a `Proxy (Component name Model Action)` instead of `Component name Model Action` as the parameter. I'm also thinking that `component_` may be the only function that really needs the concrete value for a component.

dmjio:
Thank you for this review. You are correct in that we don't need the entire value necessarily, just the component name (`Symbol`) and `action`, for `notify`. For `sample` we'd need the component name and `model`. These could be specified by the user. 

But the reason I chose to pass in the entire `component` value is that the type is inferred for you, guaranteeing correctness (the name `Symbol` and `model` / `action` must match). Otherwise, in order to ensure that the component `Symbol` name and types are aligned we'd need some global type level `Component` registry, which itself would need to be passed in via `Proxy`, and then a type level lookup would occur. This would require defining a local component and then "registering" it by placing it in a type-level `Map`.

I found passing in the value guaranteed that whatever you `sample` or `notify` is what the user actually defined. The prime variants `notify'` / `sample'` are unsafe and are used w/ dynamic components in conjunction w/ `onMountedWith` callback, or the `ask` returns the `componentId` in `Effect`, and this can be relayed to the parent (or anywhere really).

You're also correct in that circular imports are an issue. This can be seen as a benefit in that components should never be circularly used. While it is possible for `component3` to embed `component4` in each others `view`, this will lead to a runtime exception that we detect and raise during diffing. 

[...]

PR #1002, opened 2025-07-01
Add `Pub`/`Sub` system for `Component` communication 📡 

dmjio:
:artificial_satellite: Pub / Sub 📡  / Mailboxes 📬 
============

Per chats with @kaol in #1000, while `Component` are now completely isolated due to the existential type encoding, `Component` communication is not. Currently `notify` and `sample` require type-level unification on either `action` or `model`.

This has practical implications for module dependencies, and `Component` sharing, which make usage unwieldy at times (as noted in #1000)

This PR is a complete overhaul of `Component` communication, aimed at preserving isolation of `Component`, yet allowing for intra & inter-`Component` communication with `publish`, `subscribe`, `unsubscribe` primitives via a user-defined JSON protocol.

This PR aims to implement a variant of the original Elm mailbox system pioneered by @evancz. Under the hood this uses a broadcast `TChan` from `stm` (`newBroadCastTChanIO`) which only allows `dupTChan` for reads and ensures memory is reclaimed. See `Miso.Concurrent`.

`Component` can now communicate via a JSON protocol, this is embedded in the new primitives via `ToJSON` / `FromJSON` constraints. This frees up `Component` consumers to use their own types as long as the underlying `Value` is identical. This should also avoid the overhead of encoding / decoding via strings.

We can now drop `DataKinds` which was arguably a hindrance for adoption and not simple Haskell philosophy.

`Topic` are created on demand in `publish` / `subscribe` if they do not exist.

 - [x] Add `subscribe`, `unsubscribe`, `publish` primitives.
 - [x] Remove `"name"` from `Component` (all components are now dynamic)
 - [x] Rework `examples/component` to include pub sub system
 - [x] Add `stm` dep.
 - [x] Drop `notify`, `sample`.
 - [x] Update `Miso.Concurrent` to use `Topic`, a wrapper over a broadcast `TChan`.
 - [x] `componentMap` -> `components`
 - [x] Add `_id` lens (tweak `Category` instance on `Lens`)
 - [x] Update `unmount` to unsubscribe from all topics
 - [x] `defaultComponent` -> `component`
Example usage of `server` / `client` communication pattern in `examples/components`.

### DIFF
--- a/src/Miso/Internal.hs
+++ b/src/Miso/Internal.hs
@@ -1,13 +1,14 @@
 -----------------------------------------------------------------------------
-{-# LANGUAGE CPP                 #-}
-{-# LANGUAGE DataKinds           #-}
-{-# LANGUAGE LambdaCase          #-}
-{-# LANGUAGE BangPatterns        #-}
-{-# LANGUAGE KindSignatures      #-}
-{-# LANGUAGE RecordWildCards     #-}
-{-# LANGUAGE TypeApplications    #-}
-{-# LANGUAGE OverloadedStrings   #-}
-{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE CPP                        #-}
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE BangPatterns               #-}
+{-# LANGUAGE KindSignatures             #-}
+{-# LANGUAGE RecordWildCards            #-}
+{-# LANGUAGE TypeApplications           #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Miso.Internal
@@ -20,23 +21,25 @@
 module Miso.Internal
   ( -- * Internal functions
     initialize
-  , componentMap
-  , notify
-  , notify'
+  , freshComponentId
   , runView
-  , sample
-  , sample'
   , renderStyles
   , Hydrate(..)
   -- * Subscription
   , startSub
   , stopSub
+  -- * Pub / Sub
+  , subscribe
+  , unsubscribe
+  , publish
+  , Topic (..)
+  , topic
   ) where
 -----------------------------------------------------------------------------
-import           Control.Exception (throwIO)
-import           Control.Monad (forM, forM_, when, void)
+import           Control.Monad (forM, forM_, when, void, forever)
 import           Control.Monad.Reader (ask)
 import           Control.Monad.IO.Class
+import           Data.Aeson (FromJSON, ToJSON, Result, fromJSON, toJSON)
 import           Data.Foldable (toList)
 import           Data.IORef (IORef, newIORef, atomicModifyIORef', readIORef, atomicWriteIORef, modifyIORef')
 import           Data.Map.Strict (Map)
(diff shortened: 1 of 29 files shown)


dmjio:
Example usage in `examples/components/Main.hs`

```haskell
data Action
  = AddOne
  | SubtractOne
  | Mount MisoString
  | Subscribe
  | Unsubscribe
  | Notification (Result Message)
  deriving (Show, Eq, Generic)
data Message
  = Increment
  | Decrement
  deriving (Show, Eq, Generic, ToJSON, FromJSON)
main :: IO ()
main = run (startComponent server)
arithmetic :: Topic Message
arithmetic = topic "arithmetic"
-- | Demonstrates a simple server / client, pub / sub setup for 'Component'
-- In this contrived example, the server component holds the
-- incrementing / decrementing actions, and relays them to the clients
-- via the pub / sub mechanism.
--
-- Notice the server has no 'model' (e.g. `()`)
--
server :: Component () Action
server = component () update_ $ \() ->
  div_
  []
  [ "Server component"
  , button_ [ onClick AddOne ] [ "+" ]
  , button_ [ onClick SubtractOne ] [ "-" ]
  , component_ (client_ "client 1")
    [ onMountedWith Mount
    ]
  , component_ (client_ "client 2")
    [ onMountedWith Mount
    ]
  ] where
      update_ :: Action -> Effect () Action
      update_ = \case
        AddOne ->
          publish arithmetic Increment
        SubtractOne ->
          publish arithmetic Decrement
        Mount compId ->
          io_ $ consoleLog ("Mounted component: " <> compId)
        _ -> pure ()
client_ :: MisoString -> Component Int Action
client_ name = (clientComponent name) { initialAction = Just Subscribe }
clientComponent :: MisoString -> Component Int Action
clientComponent name = component 0 update_ $ \m ->
  div_
  []
  [ br_ []
  , text (name <> " : " <> ms (m ^. _id))
  , button_ [ onClick Unsubscribe ] [ "unsubscribe" ]
  , button_ [ onClick Subscribe ] [ "subscribe" ]
  ] where
      update_ :: Action -> Effect Int Action
      update_ = \case
        AddOne -> do
          _id += 1
        SubtractOne ->
          _id -= 1
        Unsubscribe ->
          unsubscribe arithmetic
        Subscribe -> do
          io_ (consoleLog "subscribing...")
          subscribe arithmetic Notification
        Notification (Success Increment) ->
          update_ AddOne
        Notification (Success Decrement) ->
          update_ SubtractOne
        Notification (Error msg) ->
          io_ $ consoleError ("Decode failure: " <> ms msg)
        _ -> pure ()
```

Zer0-:
This is really cool!

dmjio:
cc @georgefst this should let us do broadcast for free now.

dmjio:
@kaol this should make component communication loosely-coupled.

dmjio:
This is kind of like Amazon SNS, GCP Pub / Sub on the frontend.

src/Miso/Internal.hs
```diff
@@ -133,6 +134,248 @@ data ComponentState model action
...
+--
+-- @
+--
+-- @since 1.9.0.0
+topic :: MisoString -> Topic a
+topic = Topic
+-----------------------------------------------------------------------------
+mailboxes :: IORef (Map (Topic a) Mailbox)
+{-# NOINLINE mailboxes #-}
+mailboxes = unsafePerformIO $ liftIO (newIORef mempty)
```
georgefst, line 193: The use of a phantom type for topics was inspired by #1003, but I noticed when rebasing that there's a difference here in implementation. I had `mailboxes :: IORef (Map MisoString Mailbox)` instead. Using `unsafePerformIO` over a polymorphic reference is seriously dangerous in general, because it makes it easy to accidentally do unsafe coerces, as noted [in the docs](https://hackage.haskell.org/package/base-4.21.0.0/docs/System-IO-Unsafe.html#v:unsafePerformIO). But this is internal code and we seem to get away with it, so maybe it doesn't really matter!

dmjio: It's safe here in this case because the underlying type is still `MisoString` (the phantom type gets erased). In the example you linked [here](https://hackage.haskell.org/package/base-4.21.0.0/docs/System-IO-Unsafe.html#v:unsafePerformIO) `Int` is being coerced into `Char`, and this actually no longer core dumps (the docs say it will), but it prints `*`. In our use case its always `MisoString` so we're in the clear.

PR #1003, opened 2025-07-02
Type safe pub/sub

georgefst:
I think #1002 is a positive step in principle, but I'm not convinced we need to throw out so much type safety and force the user to care about decode errors etc. This is a rough sketch of some improvements.

The first commit arguably just gives the illusion of type safety, since it suffers the same "footguns" that @kaol mentions in #1000. The second is a proof-of-concept of the API I'd like to see, but to work nicely it really requires the ability to use side effects to initialise component state, e.g. by putting the `model` field in to IO/JSM. I have already argued that this would be a good thing anyway (see my complaints on Matrix a while back about building an encapsulated clock component).

Basically, the idea is that creating a topic returns an abstract identifier parameterised by the type of message it supports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,9 +161,9 @@ main = run (startComponent component)
 foreign export javascript "hs_start" main :: IO ()
 #endif
 ----------------------------------------------------------------------------
--- | `defaultComponent` takes as arguments the initial model, update function, view function
+-- | `component` takes as arguments the initial model, update function, view function
 component :: Component Model Action
-component = defaultComponent emptyModel updateModel viewModel
+component = component emptyModel updateModel viewModel
 ----------------------------------------------------------------------------
 -- | Empty application state
 emptyModel :: Model
--- a/README.md
+++ b/README.md
@@ -129,7 +129,6 @@ This file contains a simple `miso` counter application.
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards   #-}
 {-# LANGUAGE LambdaCase        #-}
-{-# LANGUAGE DataKinds         #-}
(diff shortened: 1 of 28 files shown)


dmjio:
@georgefst I did not explicitly close this, please rebase against `master` and re-open a new one if you're so inclined.

georgefst:
I'm actually inclined to keep this closed. The first commit has been merged via #1002. And the more ambitious and controversial second part doesn't seem to be to your tastes, which is fair enough. This is actually very similar in a sense to our previous discussions about whether component mounting should be an explicit side effect. It comes down to that if I were re-designing Miso from scratch I think I'd make more side-effect-y things explicit, rather than using `unsafePerformIO` internally. But there are trade-offs.

For the record, I don't actually know whether the more type-safe API is any less flexible. There definitely isn't as much of an issue with tight component coupling as there was with the old `notify`/`sample` API, because the topic used for sending messages is separable from the component itself, and can be passed in as a parameter as seen in the example. In general, I suppose the topics would be registered at the root of the subtree of components which use it, and passed around from there. This plumbing is a bit of extra effort, but I'd guess that it's not bad in practice, and that the safety makes it worth it, especially not having to worry about decode errors. There is possibly an issue with race conditions as pointed out on Matrix, but I suspect this isn't insurmountable.

And note that there are limits to the type safety provided by the simple version, in that we can easily create two topics with the same underlying string. This is why we have to keep the explicit decode error handling. Although there might be other interesting ideas in this space, like putting typereps in the map...

dmjio:
Yea I think if `Component` are to be distributed, we might see naming conventions like `package-name:component-name:version` to encode the JSON protocol version in `Topic`, similarly to what would be encountered in any JSON API.

PR #1009, opened 2025-07-03
nix: Switch from flake-utils to flake-parts

srid:
https://flake.parts/ allows us to keep our Nix modular, and paves way for using `haskell-flake`.

This PR is fairly simple since, at its most basic usage, flake-parts provides a 1-on-1 translation to flake-utils' `eachSystem` pattern.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,27 +1,33 @@
 {
   description = "A flake for miso";
 
-  inputs = {
+  nixConfig = {
+    extra-substituters = [
+      "https://haskell-miso-cachix.cachix.org"
+    ];
+    extra-trusted-public-keys = [
+      "haskell-miso-cachix.cachix.org-1:m8hN1cvFMJtYib4tj+06xkKt5ABMSGfe8W7s40x1kQ0="
+    ];
+  };
 
+  inputs = {
     nixpkgs.url =
       "https://github.com/alexfmpe/nixpkgs/archive/b594b289740a2bc917ed9c66fef5d905f389cb96.tar.gz";
 
-    flake-utils.url =
-      "github:numtide/flake-utils";
-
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -1,27 +1,33 @@
...
 
-  inputs = {
+  nixConfig = {
+    extra-substituters = [
+      "https://haskell-miso-cachix.cachix.org"
+    ];
+    extra-trusted-public-keys = [
+      "haskell-miso-cachix.cachix.org-1:m8hN1cvFMJtYib4tj+06xkKt5ABMSGfe8W7s40x1kQ0="
+    ];
+  };
```
srid, line 11: I had to move `nixConfig` to top-level since it wouldn't get used otherwise.

dmjio:
Okay this looks great but it still doesn't re-export the `ghc-wasm-meta` flake, or our cached GHCJS. I guess what benefits does this provide over vanilla flakes?

To merge the flake.nix I need the following features:

- [ ] Put `wasm32-wasi-ghc` and `javascript-unknown-ghcjs-9122` in the users `$PATH`
- [ ] Use vanilla flakes (users shouldn't need to do more than one step to acquire the cross compiler)

I do *not* expect users to know or do `nix` at all. The `nix` is purely a formality used to acquire the cross compilers (WASM / GHCJS) until GHC becomes truly retargettable via `zig cc` (or somehow), like what MicroHS already is.

srid:
Yea, but just to be clear this PR is functionally equivalent to #996 in its _current_ (draft) state. It doesn't add anything on top, except for swapping flake-utils for flake-parts and fixing the nixConfig thing.

Note that I opened the PR targetting the `flake-out` branch, not `master`.

dmjio:
Okay, sounds good .

PR #1010, opened 2025-07-04
Canvas: remove unnecessary () arguments

juliendehos:
Currently, we have something like this:
```hs
draw :: Canvas ()
draw = do
  ...
  beginPath ()
  ...
  fill
```
This PR removes the `()` argument for `beginPath`, `clip`, `save`, `restore` and `stroke` so we can use them in the same way as `fill`.


juliendehos:
Add `()` to `fill`, instead. :sweat_smile:

dmjio:
We might need to update the canvas 2d example as well

juliendehos:
I didn't find any call to `fill` in the canvas2d example. :thinking:

dmjio:
Ah it's `fillRect`, nevermind.

PR #1012, opened 2025-07-05
Add `modelCheck :: Bool` to `Component`

dmjio:
`modelCheck` is used to decide if we should diff the model before diffing the virtual DOM. By default this is `True`. It is recommended to set to `False` when developing full-screen `canvas` applications. If `False` model diffing is bypassed and the `Component` will not block on an empty event queue. This means `requestAnimationFrame` is called in a tight loop (like what `three.js` `setAnimationLoop` does).

- [x] Update `canvas2D` and three examples. 

This should make canvas applications slightly more performant.

This also means we can drop `GetTime` / `SetTime` boilerplate, since we don't need to update the `model` to force a `draw` if `modelCheck = False`.


dmjio:
Will revisit later.

PR #1023, opened 2025-07-07
Adds `.parentNode` check during delegation.

dmjio:
Each event delegator runs separately for each component. How event handling occurs is dependent on the scheduler and the synchronous callback implementation, this varies by backend (GHC / WASM / GHCJS).

In WASM, during component unmounting, event undelegation is not occuring the same way as the JS backend (due to a thread scheduling difference). Events raised from freshly unmounted components can reach `document` (this does not occur in GHCJS86/9122). This has the effect of calling `document.parentNode`, which is always is `null`.

This patch avoids that from occurring for the WASM backend.

cc @Zer0-


Zer0- (approved):
I can confirm this fixes the bug for https://github.com/Zer0-/my-miso-example/tree/parentNode_bug_report

PR #1043, opened 2025-07-16
Inline `Component`

dmjio:
# Inline `Component` 🔬 

This PR introduces aesthetic / ergonomics upgrades for `Component`, and some significant improvements to internals.

- [x] Inline `Component` into the virtual DOM on `mount()`, remove intermediates `<div>` between `Component`.

Currently there is an intermediate `<div>` that is introduced to demarcate `Component`. This `<div>` is used for delegation of events, but otherwise serves no purpose. This is undesirable for many reasons, styling, end-user confusion when encountering an element they did not define, etc. Currently the virtual DOM received from the `mount()` callback is appended to this `<div>`

This PR changes a Component's `mount()` to **overwrite** (instead of `.appendChild` to DOM / `.append` to virtual DOM) the current virtual DOM node with that received from the `syncCallback` in the Haskell layer. This stitching allows the generated `View` to have a 1:1 correlation with the physical DOM when using the `component_` feature.

- [x] Remove `[Attribute action]` from `component_`

This is a byproduct of the above (removal of intermediate `<div>`), since the `<div>` was removed, so are it's attributes.

```haskell
-- N.B. [Attribute action] removed
component_ :: Component model action -> View a
component_ = VComp . SomeComponent
```

- [x] Top-level `miso` and `startComponent` now utilize `component_` 

All `Component` are now treated identically when it comes to virtual DOM construction, including top-level components (previously was not the case). Practically this just means we use the new `component_` to wrap top-level component on application start.

- [x] `ComponentId` has been placed back into the `r` of `Effect`'s `RWS`

This is for simplicity of certain exposed primitives (don't have to fetch it from within `JSM`), it is still possible to acquire the the `Component` `DOMRef` (`componentMount`) via the internal `ComponentState`, through `getComponent` , `getParentComponent`, primitives etc. We might consider adding more to the `r` for convenience (`parentId`, etc).

- [ ] Events now bubble up between `Component`

Previously `Component` were isolated for both model state and event bubbling. Through user feedback (and inspecting other Component implementation internals) we've decided to allow events to propagate / bubble between `Component`, this should give users a more natural / seamless feel when interacting with `Component`, as would be experienced w/o `Component`. This is a quality of life / ergonomics improvement.

Users can just use `stopPropagation` on the `onWithOptions` function when dealing with events that will propagate between `Component`. This should be documented behavior though since it is not necessarily intuitive. A `propagation :: Bool` was considered (that lives in `Component`), but tabling that for now.

- [x] Mounting now occurs in the top-most node of a `Component` (previously child mounting callbacks occurred in the parent `Component`)

Again, now that the intermediate `<div>` is removed the `onMounted` callbacks occur in the top-level node of the current `Component`, it's similar to `onCreated` but only works for the top-level virtual DOM node, and occurs after the `mount()` callback, as opposed to post `createElement` call.

- [x] Embed `unmount` into `mount`. 

`unmount()` no longer lives on the same virtual DOM node as `mount()`. `mount()` generates a new virtual DOM tree where `unmount()` is set on the top-level virtual DOM node. We now rely on the GHC GC and / or the JS GC to recycle the callback for `unmount`.

Some asides, 

 - `key_` is required as a `StableName`. React has the same limitation here as `miso`. Getting around this limitation is ongoing research 🔬 (but it's not looking promising). Context below.
 
Due to the immutable nature of the virtual DOM, and the ignorance of `View` in the Haskell layer to the state of the current virtual DOM in JS and its physical mapping to the actual DOM, there is no way for us to know if a user is intending to replace an existing `Component` with a new `Component` (which would invoke the expensive `unmount()` and `mount()` operations) or if the new `Component` generated is identical to the existing `Component` and should just be copied by reference (avoiding a `unmount`).

If we choose to always `unmount () `/ `mount ()` then we run into a situation where the state is reset for `Component` where it shouldn't be necessarily. This is an unacceptable situation obviously. If we do the opposite and choose to never `unmount()` / `mount()`, then in circumstances where we want to replace one `Component` with a new `Component` it will look as if nothing happened, again unacceptable. We are currently unable to differentiate this case on behalf of the user, therefore the user must tell us (this is the current state). 

Therefore a `StableName` identifier must be introduced (ala `key_`) to differentiate between an existing `Component` and new `Component` during `View` construction. This is how React does it. `key_` is also overloaded to be used in `Component` diffing along with child list reconciliation optimizations.

If this situation changes I will make a PR, ideally end users can always be ignorant during `View` construction. 

Misc. things

- [x] `ToHtml` adjusted (don't gen `<div>`)
- [x]  `Tag` synonym intro'd

- [x] Update `component.spec.ts` test

cc @brakubraku


dmjio:
Rethinking this because,

1) Parent components need child component ID when they mount, current setup allows for that
2) The intermediate `<div>` really isn't a problem, most CSS resets remove all styling anyways
3) Keeping the `[Attribute action]` list allows us to easily specify `key_`, in an obvious intuitive way
4) Mutating the virtual DOM hierarchy is a bad idea (should stay immutable)

Closing for now, will revisit.

PR #1054, opened 2025-07-25
Moves SSE to client-only applications

dmjio:
This serves as a precursor to splitting miso-examples out into separate repos. cc @juliendehos.

This drops the server component of the SSE example, removes nginx hosting, and consumes an example endpoint:
  - https://echo.websocket.org/.sse


dmjio:
Closing in favor of https://github.com/haskell-miso/miso-sse per @juliendehos migration work

PR #1059, opened 2025-07-26
Add `componentParentId` and `ComponentInfo`.

dmjio:
This parameterizes `Effect` by `ComponentInfo`, allowing users to extract a parent's ComponentId (if it exists), for message passing. The parent ID is set during Component initialization by tree climbing w/ `miso.getParentComponentId`.

This makes it so both component ID and component mount DOMRef are avaialable to end users in `Effect` via `ask`, `asks`, or `view`.

### DIFF
--- a/src/Miso/Effect.hs
+++ b/src/Miso/Effect.hs
@@ -25,6 +25,9 @@ module Miso.Effect
   , Sub
   , Sink
   , DOMRef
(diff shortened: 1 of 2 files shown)


dmjio:
Subsumed by https://github.com/dmjio/miso/pull/1047

Issue #1063, opened 2025-07-27
Root component can't subscribe to topics.


marmay:
I don't know, whether this is intentional, but the root component does not seem to be able to subscribe to topics. I get an error during runtime:

    ...: IntMap.!: key -9223372036854775808 is not an element of the map
    CallStack (from HasCallStack):
      error, called at libraries/containers/containers/src/Data/IntMap/Internal.hs:601:17 in containers-0.7-   inplace:Data.IntMap.Internal
    HasCallStack backtrace:
      collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:169:13 in ghc-internal:GHC.Internal.Exception
      toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:204:5 in ghc-internal:GHC.Internal.Exception
      error, called at libraries/containers/containers/src/Data/IntMap/Internal.hs:601:17 in containers-0.7-inplace:Data.IntMap.Internal`

When I wrap my root component by replacing `startComponent` with the following code, everything seems to work as expected:

    runApp c = startComponent $ component () (const $ pure ())
                                                     $ \() -> div_ [] +> c`

I don't know, whether this is intentional, but at least, it is highly unexpected.

dmjio:
Not intentional, currently working on a testsuite for `Miso.Runtime`, this will be covered.

dmjio:
I think what is going on here is that the top-level `Component` doesn't call `mount()` explicitly, we do it in the Haskell layer. This means the `componentId` wasn't getting set for top-level `Component`. I moved the `componentId` setting out of the JS `mount()` callback and into the `iniitalize` function. So setting the `componentId` should now be uniform for both top-level `Component`, and all subsequent `Component` in descendants.

So the error below

```
...: IntMap.!: key -9223372036854775808 is not an element of the map
```

Means that the call to `FFI.getComponentId` was returning garbage, and therefore the `IntMap` lookup was failing.

I tested #1064 on the pubsub example and the parent was able to subscribe to the topic, without needing to be nested. If you'd like, you could test the hash on your own example @marmay 

ddb6f547a160131488b18077e38565b753d82ecf

dmjio:
https://github.com/dmjio/miso/pull/1064

PR #1064, opened 2025-07-28
Set `componentId` on initialization.

dmjio:
- [x] Sets `componentId` in initialize.
- [x] `componentMount` -> `componentDOMRef`
- [x] `JSVal` -> `DOMRef`
- [x] Move `componentId` setting out of JS + regen JS


dmjio:
This addresses #1063

PR #1070, opened 2025-08-01
Drop get* variant functions, prefer `parent`, use `ask`.

dmjio:
These functions are no longer necessary because we put `ComponentId` in the `Reader` for `Effect` and the new `parent` function in #1047 will subsume `getParentComponentId` (and further it uses the `FFI.Internal` variant).


dmjio:
This also fixes some issues with how the FFI module is haddocked

PR #1079, opened 2025-08-03
Introduce data bindings into `Component`.

dmjio:
This allows the children to bind to a parent's field (via a `Lens`) and "react" to changes. ~~Right now just unidirectional, could be extended to bidirectional.~~ (*edit*: also bidirectional)

This is equivalent to [React props](https://react.dev/learn/passing-props-to-a-component), and is the primary mechanism by which components in React communicate.

We can implement combinators for bidirectional or unidirectional `Component` communication.

Bidirectional could allow sibling communication (`<-->`) by way of the `parent`.

```haskell
(-->), (<--), (<-->) :: Lens parent field -> Lens child field -> Binding parent field child
```

- [x] Ensure that `Direction` is respected when setting up bindings

### DIFF
--- a/src/Miso/Runtime.hs
+++ b/src/Miso/Runtime.hs
@@ -176,14 +176,6 @@ initialize Component {..} getView = do
   _ <- FFI.forkJSM eventLoop
   pure vcomp
 -----------------------------------------------------------------------------
--- | dmj: we only want to call this if props have been defined
--- filter this function on 'ChildToParent'
---
--- What we do here is fork a thread on current ComponentDiffs
--- Then iterate on props, see if there is a ChildToParent, if so
--- then we grab the parent, read the child from the lens, set it on the parent
--- then we call 'serve' on the parent
---
 synchronizeChildToParent
   :: DOMRef
   -> IORef model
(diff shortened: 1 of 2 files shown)


src/Miso/Runtime.hs
```diff
@@ -102,36 +105,163 @@ initialize Component {..} getView = do
...
+      -- dmj: impossible case, parent always mounted
+      pure Nothing
+    Just parentId -> do
+      IM.lookup parentId <$> liftIO (readIORef components) >>= \case
+        Nothing -> do
+          -- dmj: another impossible case, parent always mounted
+          pure Nothing
+        Just parentComponentState -> do
+          bindProperty parentComponentState
+          -- dmj: ^ assume parent state on initialization
```
dmjio: should be "parent assumes child state on init"

src/Miso/Runtime.hs
```diff
@@ -102,36 +105,163 @@ initialize Component {..} getView = do
...
       message <- liftIO (readMail =<< copyMailbox componentMailbox)
       mapM_ componentSink (mailbox message)
-  let vcomp = ComponentState {..}
+
+  -- bindings
+  componentParentToChildThreadId <-
+    synchronizeParentToChild
+      componentDOMRef
+      componentModelNew
+      bindings
```
dmjio:
```haskell
[ binding | Binding dir _ _ <- bindings, dir `elem` [ Bidirectional, ParentToChild ] ]
```

PR #1081, opened 2025-08-04
Use `title_` from `Element`, as opposed to `Property`.

dmjio:
We need this for the haskell-miso.org repo.

- [x] Prioritizes `<title>` element over property


dmjio:
Solved in https://github.com/haskell-miso/haskell-miso.org/commit/2682fd366a3919062052e067e5c8770bf3a8dd33

PR #1090, opened 2025-08-07
Stronger lens agnosticism

georgefst:
A follow-up of sorts to #1088 ("lens atheism"?).

After using the new bindings stuff for a couple of days, I've come up with some slight modifications which IMO make for a cleaner API. My main motivation is creating a `miso-optics` library that's easy to use without fiddling with import hiding. This is just a suggestion, and thus code is in a messy PoC form.

With these changes:

- Users who don't want lenses at all can just not import `Miso.Lens`. The non-lens `<-->` is a bit unwieldy, but we could add a suggestion in the Haddocks saying "look at this other version which is more convenient, if you're comfortable with lenses". Anyway, presumably it's seen as less important than the unidirectional ones since it was removed completely for a while.
- Users who just want to use Miso lenses can import `Miso.Lens` and `Miso.Lens.Binding`. The latter would probably be better off in `Miso.Lens` itself, and I only haven't done that yet because it would require some slight rejigging to avoid cyclic modules.
- Users who want to use Van Laarhoven lenses can import `Miso.Lens.VL` for helpers and conversions from `lens`/`microlens`. We don't even really need to expose `lensFromVL` etc. given that bindings are the only part of the Miso API that actually uses a lens interface (we probably don't need conversions the other way regardless). In fact, `Miso.Lens.VL` could be a separate library, much like `miso-optics`, unless we want to hide the internals of Miso lenses (which maybe we should).
- Users who want to use `optics` (me), can depend on a `miso-optics` library. This doesn't exist yet, but the code would look very similar to `Miso.Lens.VL`.

One thing I haven't done here but would strongly consider is re-inlining the type synonyms for `Getter` and `Setter`. Without inheritance like the proper ones, they're not very useful and it feels like a bit of a false analogy.

As an alternative to all this, we _could_ drop `Miso.Lens` entirely in favour of `optics-core`, which would result in 4 new transitive dependencies, none of them big and all well-maintained...

### DIFF
--- a/miso.cabal
+++ b/miso.cabal
@@ -137,6 +137,8 @@ library
     Miso.Html.Event
     Miso.Html.Property
     Miso.Lens
+    Miso.Lens.Binding
+    Miso.Lens.VL
     Miso.Mathml
     Miso.Mathml.Element
     Miso.Mathml.Property
--- a/src/Miso/Lens.hs
+++ b/src/Miso/Lens.hs
@@ -147,21 +147,14 @@ module Miso.Lens
   , this
   -- *** Re-exports
   , compose
-  -- *** Conversion
-  , Lens'
-  , toVL
-  , fromVL
   ) where
 ----------------------------------------------------------------------------
 import Control.Monad.Reader (MonadReader, asks)
 import Control.Monad.State (MonadState, modify, gets)
-import Control.Monad.Identity (Identity(..))
 import Control.Category (Category (..))
 import Control.Arrow ((<<<))
-import Data.Functor.Const (Const(..))
 import Data.Function ((&))
 import Data.Functor((<&>))
-import Data.Kind (Type)
 import Prelude hiding ((.))
 ----------------------------------------------------------------------------
 import Miso.Util (compose)
@@ -189,22 +182,6 @@ type Getter record field = record -> field
 ----------------------------------------------------------------------------
 -- | Type to express a setter on a @Lens@
 type Setter record field = field -> record -> record
-----------------------------------------------------------------------------
--- | Van Laarhoven formulation, used for conversion w/ 'miso' @Lens@.
-type Lens' s a = forall (f :: Type -> Type). Functor f => (a -> f a) -> s -> f s
-----------------------------------------------------------------------------
--- | Convert from `miso` @Lens@ to Van Laarhoven @Lens'@
-toVL :: Lens record field -> Lens' record field
-toVL Lens {..} = \f record -> flip _set record <$> f (_get record)
-----------------------------------------------------------------------------
--- | Convert from `miso` @Lens@ to Van Laarhoven @Lens'@
-fromVL
-  :: Lens' record field
-  -> Lens record field
-fromVL lens_ = Lens {..}
-  where
-    _get record = getConst (lens_ Const record)
-    _set field = runIdentity . lens_ (\_ -> Identity field)
 ----------------------------------------------------------------------------
 -- | Lens are Categories, and can therefore be composed.
 instance Category Lens where
--- a/src/Miso/Lens/Binding.hs
+++ b/src/Miso/Lens/Binding.hs
@@ -1,0 +1,11 @@
+module Miso.Lens.Binding where
+
+import Miso.Lens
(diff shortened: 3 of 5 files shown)


dmjio:
@georgefst glad to see this, I think a `miso-optics` adaptor package makes sense. As you've seen, I've gone ahead and moved `Binding` from

(before)
```haskell
data Binding parent child = forall field . Binding (Lens parent field) (Lens child field)
```

(after)
```haskell
data Binding parent child
  = forall field . ParentToChild (Getter parent field) (Setter child field)
  | forall field . ChildToParent (Setter parent field) (Getter child field)
  | forall field . Bidirectional (Getter parent field) (Setter parent field) (Getter child field) (Setter child field)
```

The latter allows us now to use any lens lib (be it `microlens`, `optics`, `Miso.Lens`, etc.).

I've gone ahead and reinstated the bidirectional combinators for Van Laarhoven and miso's simple lens, implemented in terms of `Binding`.

```haskell
-- | Bidirectionally binds a child field to a parent field, using @Lens@
--
-- This is a bidirectional reactive combinator for a miso @Lens@.
--
-- @since 1.9.0.0
(<-->) :: Lens parent field -> Lens child field -> Binding parent child
p <--> c = Bidirectional (_get p) (_set p) (_get c) (_set c)
-- | Bidirectionally binds a child field to a parent field, using @Lens'@
--
-- This is a bidirectional reactive combinator for a Van Laarhoven @Lens'@
--
-- @since 1.9.0.0
(<--->) :: Lens' parent field -> Lens' child field -> Binding parent child
p <---> c = Bidirectional (_get (fromVL p)) (_set (fromVL p)) (_get (fromVL c)) (_set (fromVL c))
```

and also altered the `-->` , `<--` unidirectional combinators to only take `Getter` /  `Setter` per your recommendation.

```haskell
-- | Type to express a getter on a @Lens@
type Getter record field = record -> field
-- | Type to express a setter on a @Lens@
type Setter record field = field -> record -> record
-- | Unidirectionally binds a parent field to a child field, using @Getter@ and @Setter@
--
-- @since 1.9.0.0
(-->) :: Getter parent a -> Setter model a -> Binding parent model
(-->) = ParentToChild
-- | Unidirectionally binds a child field to a parent field, using @Getter@ and @Setter@
--
-- @since 1.9.0.0
(<--) :: Setter parent a -> Getter model a -> Binding parent model
(<--) = ChildToParent
```

So the usage looks something like:

```haskell
 { bindings =
      [ view childCounter --> set x
      , set childCounter <-- view x
      ]
  }
 { bindings =
      [ (^. childCounter) --> (x .~)
      , (childCounter .~) <-- (^. x)
      ]
  }  
  { bindings =
      [ _get childCounter --> _set x
      , _set childCounter <-- _get x
      ]
  }
  { bindings =
      [ _childCounter --> \field record -> record { _x = field }
      , (\field record -> record { _childCounter = field }) <-- _x
      ]
  }
    { bindings =
      [ childCounter <--> x
      ]
  }
  { bindings =
      [ childCounter <---> x
      ]
  }
```

Should all be valid...

So regarding `miso-optics`, @georgefst , I think we could make a package that exposes an implementation like:

```haskell
(<--->) :: Optic' k is parent field -> Optic' k is child field -> Binding parent child
parent <---> child = Bidirectional (^. parent) (parent .~) (^. child) (child .~)
```

The same could be done for the unidirectional lenses as well.

I'd prefer we keep `Miso.Lens` as a gateway into basic lensing (with a seatbelt), then users could learn all the fanciness that comes with generalizing to VL at their own pace.

Also, `miso-lens` or `miso-vl` shouldn't be required.

dmjio:
https://github.com/dmjio/miso/pull/1111

dmjio:
https://github.com/dmjio/miso/pull/1112

dmjio:
@georgefst  A `miso-optics` lib (where `optics` is the lens lib) is still welcome.

```haskell
(<--->) :: Optic' k is parent field -> Optic' k is child field -> Binding parent child
parent <---> child = Bidirectional (^. parent) (parent .~) (^. child) (child .~)
```

PR #1094, opened 2025-08-08
Add `onContextMenuWithOptions` support

dmjio:
This disables right-click context menu from appearing when used w/ `preventDefault`

```haskell
div_ [ onContextMenuWithOptions NoOp defaultOptions { preventDefault = False } ] [ ]
```


dmjio:
cc @juliendehos

PR #1098, opened 2025-08-08
Add `wai-middleware-static` to `Miso.Run` debug `Middleware`.

dmjio:
This means that usage of `CSS.url "mario.png"` w/ `Miso.Style` will "just work" when developing from GHCi. Assuming `'mario.png'` is in the same directory where GHCi is running.


dmjio:
tested this locally and it works

dmjio:
image

dmjio:
Should upstream this into https://github.com/haskell-miso/jsaddle-run

PR #1104, opened 2025-08-11
Add `Miso.Subscription.Util.createSub`

dmjio:
This factors out the resource finalization of `createSub` into a `Utils` module so other `Sub` can benefit from it.

- [x] Refactor `keyboardSub`, `sseSub` to use resource finalization from `createSub`
- [x] Re-export `JSM`


dmjio:
This also has the byproduct of allowing the `Keyboard` and `SSE` subscriptions to be dynamic (not just live in top-level `subs`)

PR #1107, opened 2025-08-13
Properly handle boolean HTML attributes

dmjio:
Remove `disabled` altogether from attribute list if removed on new virtual DOM. Before it was just setting its value to an empty string.

https://html.spec.whatwg.org/#boolean-attributes


dmjio:
https://meiert.com/blog/boolean-attributes-of-html/

PR #1124, opened 2025-08-20
Add `connect*` variants for `EventSource`.

dmjio:
This is identical to https://github.com/dmjio/miso/pull/1123, but for `EventSource`.

SSE is a text protocol, so only JSON / Text we support variants for.


dmjio:
tested both `connectJSON` / `connectText` on `miso-sse`, and it works as expected

PR #1125, opened 2025-08-20
Add content-specific `fetch` variants.

dmjio:
Instead of a single `fetch`, add small utility functions for web requests. Flesh-out on an as-needed basis.

In step with https://github.com/dmjio/miso/pull/1124 and https://github.com/dmjio/miso/pull/1123, this adds utility functions, matching GET, POST, PUT verbs with content types text, json, arrayBuffer and blob data.


dmjio:
We still need to add support for more content types like `FormData`, `URLSearchParams`, `DataView`, `File`, etc.

PR #1137, opened 2025-08-25
Define `hex` in terms of `MisoString`.

dmjio:
Current `hex` is broken, things like `hex 0x000ccc` won't work (0-padding is stripped). Further, we can't just prepend the 0's because hex rules are different when using 3-digit hex vs. 6-digit hex, and we can't assume the user's intention.

Therefore, we will define `hex` in terms of `MisoString` and recommend its use with `OverloadedLabels`, which is identical to CSS syntax.

```haskell
grey :: Color
grey = hex #ccc
```


dmjio:
@juliendehos

PR #1161, opened 2025-09-07
Add `componentParentId`, `rootComponentId`.

dmjio:
This is necessary for native development API calls.

- [x] Add `componentParentId` to `ComponentState`
- [x] Adds `rootComponentId`


dmjio:
This was tested against the `miso-reactive` examples, as well as the `miso-websocket` example which both make use of `componentParentId`.

PR #1171, opened 2025-10-05
adds utility functions for JSM in Effect

theNerd247:
adds modifyJSM, beforeAll, and afterAll

### DIFF
--- a/src/Miso/Effect.hs
+++ b/src/Miso/Effect.hs
@@ -224,7 +224,7 @@ beforeAll = modifyAllJSM . (*>)
 -- Example usage:
 --
 -- > -- log that a given effect completed execution
--- > beforeAll $ consoleLog "Done!"
+-- > afterAll $ consoleLog "Done!"
 afterAll :: JSM () -> Effect parent model action -> Effect parent model action
 afterAll = modifyAllJSM . (<*)
 -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dmjio:
LGTM 👌🏼 , maybe we could add a `before`, `after` too.

```haskell
before :: JSM () -> Effect parent model action
before action = censor (\actions -> const action : actions)

after :: JSM () -> Effect parent model action
after action = censor (\actions -> actions ++ [const action])
```

theNerd247:
Your definition of `before` is isormorphic to `io_` (`before a x ~ io_ a *> x`)

The same is for `after`: `after a x ~ x *> io_ a` 

Maybe it would be better to add comments to `beforeAll` and `afterAll` to show the user how to do this instead? I feel like introducing `before` and `after` might lead the user away from the intended semantics of using `Effect`

dmjio:
You're right, and my definition won't actually affect any pre-existing `JSM` action. Each element in the list gets its own thread, so the `threadDelay` example you have above wouldn't work as expected with my above `before` definition (since the `threadDelay` would occur on a different thread, it wouldn't actually delay the Websocket connection). 

But in the case of `beforeAll`, the `threadDelay` would get applied to all `JSM` actions in the list, which might also be unexpected. What I'd like is to just apply the `threadDelay` to the first `JSM` element in the list.

```haskell
before :: JSM () -> Effect parent model action
before action = censor $ \case
  [] -> []
  x : xs -> (\sink -> action >> x sink) : xs
```

dmjio:
We could add `before`, and `after` in a separate PR though. I'm fine merging the above now as is, if you're ready.

theNerd247:
Got it. Yeah I can do that in another PR. I'm good to merge now.

PR #1172, opened 2025-10-06
Adds `initialModel :: Maybe (JSM model)` to `Component`

Zer0-:
This initialModel action is run before a component is rendered to give every component a chance to initialize its own state.


dmjio (changes requested):
Can we move to calling it `hydrateModel` and pass `URI` as a param, per your recommendation earlier

Zer0-:
> Can we move to calling it `hydrateModel` and pass `URI` as a param, per your recommendation earlier

I think these are addressed.

`data URI` and friends had to be moved into Types.hs, since the URI type comes from Router, and Router imports Types, it would have caused a circular import.

dmjio (approved):

PR #1173, opened 2025-10-10
Wire-up `hydrateModel` to `ToHtml` instance.

dmjio:
During server pre-rendering we'd like to invoke `hydrateModel`. This allows HTML to be generated with desired model state. Since adding `Component` , hydration is now required for all nested `Component`. It can be unwieldy to thread model state through to all sub `Component` in the `ToHtml` instance (this is what `hydrateModel` was introduced to ameliorate - while simultaneously preserving `Component` encapsulation as enforced by an existential barrier). 

This change allows model hydration on the server when prerendering, so the `ToHtml` `Component` instance can generate HTML in the same way that the client can generate the virtual DOM, facilitating pointer copying between the two, and ultimately hydration of `View`.

- [x] Oscillate `JSM` / `IO` when using `jsaddle` or vanilla GHC w/o `jsaddle`
- [x] Invoke `hydrateModel` from within `toHtml` when pre-rendering on the server (w/o `jsaddle`)
- [x] Adds `jsaddle` flag.

### DIFF
--- a/src/Miso/Html/Render.hs
+++ b/src/Miso/Html/Render.hs
@@ -157,7 +157,7 @@ toHtmlFromJSON (Array a)    = fromMisoString $ ms (show a)
 -- | Used for server-side model hydration, internally only in 'renderView'.
 --
 -- We use 'unsafePerformIO' here because @servant@'s 'MimeRender' is a pure function
--- yet we need to allow the user's to hydrate in 'IO'.
+-- yet we need to allow the users to hydrate in 'IO'.
 --
 getInitialComponentModel :: Component parent model action -> model
 getInitialComponentModel Component {..} =


src/Miso/Html/Render.hs
```diff
@@ -41,8 +47,19 @@ instance ToHtml [View m a] where
...
 ----------------------------------------------------------------------------
 -- | Render a @Component parent model action@ to a @L.ByteString@
+#ifndef JSADDLE
 instance ToHtml (Component parent model action) where
-  toHtml Component {..} = renderView (view model)
+  toHtml Component {..} = do
+    case hydrateModel of
+      Nothing ->
+        renderView (view model)
+      Just action -> unsafePerformIO $ do
```
Zer0-: Is this legal?

src/Miso/Types.hs
```diff
@@ -83,6 +87,10 @@ import           Data.String (IsString, fromString)
 import qualified Data.Text as T
 import           Language.Javascript.JSaddle (ToJSVal(toJSVal), Object(..), JSM)
 import           Prelude
+#ifndef JSADDLE
+import           System.IO.Unsafe (unsafePerformIO)
+import           Control.Exception (catch, SomeException)
```
Zer0-: Also need `SomeException` client-side to compile Miso. This block is server-side only.

src/Miso/Types.hs
```diff
@@ -83,6 +87,10 @@ import           Data.String (IsString, fromString)
 import qualified Data.Text as T
 import           Language.Javascript.JSaddle (ToJSVal(toJSVal), Object(..), JSM)
 import           Prelude
+#ifndef JSADDLE
+import           System.IO.Unsafe (unsafePerformIO)
```
Zer0-: Same with unsafePerformIO, needed client-side and server-side to compile

src/Miso/Types.hs
Zer0-: Same with catch, needed client-side and server-side to compile

Zer0-:
[This](https://github.com/dmjio/miso/pull/1188) implements all of the fixes to building I just mentioned, <del>we can fix building client-side and merge this.</del>

edit: this branch crashes my app even with the fixes, with a generic client-side message. I want to fix this to merge it

PR #1189, opened 2025-10-27
Add `postJSON'` to both send and receive JSON

dmjio:
Introduced a new `postJSON'` function for handling JSON `POST` requests with enhanced error handling.


dmjio:
@CodingArdor

PR #1201, opened 2025-11-01
fix renderScripts for ImportMap

fanshi1028:
[Importmap ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap#bare_modules)


dmjio:
Thank you for the fix @fanshi1028 !

Am impressed with your advanced usage of `miso`. If you'd like to join our matrix please do.

https://matrix.to/#/#haskell-miso:matrix.org

PR #1202, opened 2025-11-01
Properly escape `HTML` on the server.

dmjio:
When serializing templates on the server using `text` we now automatically escape HTML (this is when the `jsaddle` flag is not in use).

The only exception is when serializing `script_` contents. For that we introduce the `textRaw` combinator which doesn't escape HTML.

- [x] Use `htmlEncode` in `text`, move to `Miso.Types`
- [x] Expose `textRaw` and use in `script_`
- [x] `CPP` around `text` w/ `htmlEncode`


Zer0- (approved):
This looks like we got it right this time

PR #1204, opened 2025-11-04
Improve haddocks for style_ and app functions

jhrcek:


src/Miso/Types.hs
```diff
@@ -221,7 +221,9 @@ data ROOT
 -- | 'Eq' instance for 'ROOT'
 instance Eq ROOT where _ == _ = True
 -----------------------------------------------------------------------------
--- | For top-level t'Miso.Types.Component', 'ROOT' must always be specified for parent.
+-- | A miso applicatoin is a top level t'Miso.Types.Component', which has no parent.
+-- This is enforce by specializing the 'parent' type parameter to 'ROOT'.
```
dmjio: enforced*

jhrcek:
```suggestion
-- This is enforced by specializing the 'parent' type parameter to 'ROOT'.
```

PR #1205, opened 2025-11-04
Improve Media haddocks

jhrcek:
While working on this (I just copy pasted the first sentence for each property/method from "Definition and Usage" section in w3 schools - because I hate those "naked" URLs in haddocks), I realized there's probably an API gap in that many of the properties are not just for getting the current value, but also for setting.

But the miso functions seems to be only usable for getting..?
Search for "sets or returns" in this PR and let me know what you think - maybe we should split the properties that have both get and set functionality into `get<property>` and `set<property>` functions?

### DIFF
--- a/src/Miso/Media.hs
+++ b/src/Miso/Media.hs
@@ -62,7 +62,7 @@ import           Miso.String
 newtype Media = Media JSVal
   deriving (ToJSVal)
 -----------------------------------------------------------------------------
--- | https://www.w3schools.com/tags/av_prop_networkstate.asp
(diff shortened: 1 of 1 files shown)


dmjio (approved):
Setting properties for things like `src_` etc. on `<video>` /  `<audio>` objects can be done with the property setting mechanism. So we only worry about getting here, and let the `view` DSL work its magic and give us the setting for free. @juliendehos has examples of this in his [audio](https://github.com/haskell-miso/miso-audio) and [video](https://github.com/haskell-miso/miso-video) projects

PR #1217, opened 2025-11-16
`Component` mount / unnmount patch (PR 2/3)

dmjio:
This is step `2/3` for `miso-lynx` readiness to consume native device APIs and utilize both JS interpreters.

- [x] Thread `ComponentId` through diffing
- [x] Modify `AppendChild` patch to account for DOM "stitching"

This occurs during `Component` mounting to ensure the parent `Component` is patched correctly to the `Component` child. We introduce a `parentComponentId` in the `AppendChild` (nullable) to facilitate this.

- [x] Adjust context to account for `Component`
- [ ] Add mounting / unmounting patch tests


dmjio:
Closing in favor of https://github.com/dmjio/miso/pull/1254

PR #1225, opened 2025-11-21
Make event delegation global, refactor mounting.

dmjio:
Currently event delegation is Component local. This change is to refactor delegation to operate on the entire Component tree. This is what React does as well.

This change also refactors `VComp` to be a virtual node to avoid introducing intermediate DOM nodes when nesting Component.

```haskell
view :: m -> View m a
view m = div_ [ id_ "foo" ] +> component (0 :: Int) noop $ \x ->
  text ("number: " <> ms x)
```

to be

```haskell
view :: m -> View m a
view m = mount [ key_ "foo" ] $ component (0 :: Int) noop $ \x ->
  text ("number: " <> ms x)
```

The above avoids introducing an intermediate `<div></div>`.

- [x] Refactor `+>` into `mount`, make `VComp` virtual
- [x] Introduces a global Component tree, and globally delegated events.
- [x] Add `componentEvents` to `ComponentState`.


dmjio:
Closing in favor of https://github.com/dmjio/miso/pull/1254

PR #1228, opened 2025-11-23
Drops `shouldSync` property, determine child sync. optimizations in `diff`

dmjio:
The `shouldSync` property was introduced as a preprocessing step to avoid logic in JS relating to determining if child synch. optimizations should be entered. Unfortunately, this "optimization" introduces a lot of additional allocations in the Haskell heap, as detected by @Zer0- . 

This is because in order to determine `shouldSync`, we have to treat linked lists as data structures as opposed to control structures, this breaks streaming into FFI functions for `VTree` construction in `JSM`. This change removes the `shouldSync` property from `VTree` and introduces a function to determine dynamically (in constant space) in JS.


ts/miso/dom.ts
```diff
@@ -175,8 +175,23 @@ function diffCss<T>(cCss: CSS, nCss: CSS, node: T, context: DrawingContext<T>):
...
+        }
+    }
+    for (var i = 0; i < n.children.length; i++) {
+        if (n.children[i].key === null || n.children[i].key === undefined) {
+            return false;
+        }
+    }
+    return true;
+}
+
```
dmjio, line 192: This is the dynamic check for child sync optimizations

PR #1232, opened 2025-11-24
Bump nixpkgs hash

dmjio:


Reijix:
Since this failed with `No space left on device`, I'd suggest looking into
https://github.com/wimpysworld/nothing-but-nix
I imagine that large nix builds will be a persistent problem for this project.

Also if you choose to use it, be sure to also follow
https://github.com/wimpysworld/nothing-but-nix?tab=readme-ov-file#grant-user-ownership-of-nix-nix-permission-edict-%E2%80%8D%EF%B8%8F
I believe that the nix runner that is used in miso needs it.

dmjio:
https://github.com/dmjio/miso/pull/1256

dmjio:
Seems like headless browser derivation is broken in this chosen nixpkgs

```
1 | const { chromium } = require('playwright');
2 | 
3 | (async () => {
4 |   const browser = await chromium.launch({ headless: true });
                                     ^
error: launch: Executable doesn't exist at /nix/store/7m2pyxfb9ym4ib5i10qy9cw4mc6x23yf-playwright-browsers/chromium_headless_shell-1179/chrome-linux/headless_shell
```

dmjio:
We'll come back to this later

PR #1238, opened 2025-11-27
Avoid expensive Map lookup

dmjio:
Gaurd the `Map` lookup behind the debug check on `logLevel`


dmjio:
This is safe to do because we check if the event exists already on the JS side, so this avoids superflous checks.

PR #1245, opened 2025-11-28
Expose HTMLInputElement methods

Reijix:
Hiya!
For my personal project I need some methods of HTMLInputElement, namely
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange

I started implementing `select` by copying the existing implementation of `focus`. The only difference is that I used `document.querySelector()` instead of `document.getElementById()`, since the result of the former can be casted to `HTMLInputElement`, while the latter always returns `HTMLElement`.

Next, I'll do the same for `setSelectionRange`, but first is there any feedback? Are these methods even welcome in miso?

### DIFF
--- a/src/Miso/FFI/Internal.hs
+++ b/src/Miso/FFI/Internal.hs
@@ -508,7 +508,7 @@ focus x = void $ jsg "miso" # "callFocus" $ (x, 50 :: Int)
 --
 -- Analogous to @document.getElementById(id).blur()@
 blur :: MisoString -> JSM ()
-blur x = void $ jsg "miso" # "callBlue" $ (x, 50 :: Int)
+blur x = void $ jsg "miso" # "callBlur" $ (x, 50 :: Int)
 -----------------------------------------------------------------------------
 -- | Fails silently if the element is not found.
 --
--- a/src/Miso/FFI.hs
+++ b/src/Miso/FFI.hs
@@ -23,8 +23,9 @@ module Miso.FFI
     -- ** DOM
   , getElementById
   , focus
-  , select
   , blur
+  , select
+  , setSelectionRange
   , alert
     -- ** Styles
   , addStyle


ts/miso/util.ts
```diff
@@ -12,6 +12,14 @@ export function callFocus(id: string, delay: number): void {
   delay > 0 ? setTimeout(setFocus, delay) : setFocus();
 }
 
+export function callSelect(id: string, delay: number): void {
+  var setSelect = function () {
+    var e : HTMLInputElement = document.querySelector('#' + id);
```
jhrcek: Maybe use getElementById and get rid of the concatenation?

```suggestion
    var e : HTMLInputElement = document.getElementById(id);
```

Reijix: Ah, I've tried that (see my explanation above), but at least in vscode this throws an error, since `getElementById` returns a `HTMLElement`, not a `HTMLInputElement`. 

However, compilation still succeeds when using `getElementById`. So maybe locally I am missing some typescript config, or have a different version. I've never worked with typescript before, so does anyone now how to fix that locally?

src/Miso/FFI/Internal.hs
```diff
@@ -508,6 +508,16 @@ focus x = do
 -----------------------------------------------------------------------------
 -- | Fails silently if the element is not found.
 --
+-- Analogous to @document.querySelector('#' + id).select()@.
+select :: MisoString -> JSM ()
+select x = do
+  moduleMiso <- jsg "miso"
+  el <- toJSVal x
+  delay <- toJSVal (50 :: Int)
+  void $ moduleMiso # "callSelect" $ [el,delay]
```
dmjio: Should be able to simplify this via the tuple `MakeArgs` instance. Other FFI should be doing this as well, to avoid `toJSVal` everywhere. 

```haskell
select :: MisoString -> JSM ()
select x = void $ jsg "miso" # "callSelect" $ (x, 50 :: Int)
 ```

File: ts/miso/util.ts [hunk 3]
```diff
@@ -12,6 +12,14 @@ export function callFocus(id: string, delay: number): void {
...
+    if (e && e.select) e.select();
```
dmjio: As @jhrcek mentions, `getElementById` will be a faster lookup. The TypeScript type hierarchy is very rigid for DOM elements. We should be able to `as`-cast into `HTMLInputElement` and use a dynamic property check to avoid issues. Below should act as a safe runtime check, but also satisfy the typechecker.

```typescript
if (e && typeof e['select'] === 'function') (e as HTMLInputElement).select();
```

Also note the `#` prefix will need to be dropped above when using `getElementById`.

src/Miso/FFI/Internal.hs
```diff
@@ -508,6 +508,16 @@ focus x = do
 -----------------------------------------------------------------------------
 -- | Fails silently if the element is not found.
 --
+-- Analogous to @document.querySelector('#' + id).select()@.
+select :: MisoString -> JSM ()
+select x = do
```
dmjio: This will need to be exported.

ts/spec/util.spec.ts
```diff
@@ -17,14 +17,21 @@ afterEach(() => {
...
+    child['value'] = 'bar';
     document.body.appendChild(child);
     callFocus('blah', 0); /* missing case */
     callFocus('foo', 0); /* found case */
     callFocus('foo', 1); /* found case */
     expect(document.activeElement).toEqual(child);
+    callSelect('blah', 0);
+    callSelect('foo', 0);
+    callSelect('foo', 1);
+    var e : HTMLInputElement = document.querySelector('#foo');
```
dmjio, line 32: Fine to use `querySelector` here in the tests.

src/Miso/FFI/Internal.hs
```diff
@@ -508,6 +508,16 @@ focus x = do
...
+  moduleMiso <- jsg "miso"
+  el <- toJSVal x
+  delay <- toJSVal (50 :: Int)
+  void $ moduleMiso # "callSelect" $ [el,delay]
+-----------------------------------------------------------------------------
+-- | Fails silently if the element is not found.
+--
 -- Analogous to @document.getElementById(id).blur()@
```
dmjio: Can we flip these two comments, so the one you added is below `Analogous to @document.getElementById(id).blur()@`.

dmjio (changes requested):
This is a welcome addition, functions that live on DOM elements are awkward to deal with, like `getBoundingClientRect()`, `focus()`, `blur()` etc. 

Also, note that you can invoke `select()` on a `"change"` event for an `<input>` if you use with `with*` variants of event handlers.

https://haddocks.haskell-miso.org/miso/Miso-Html-Event.html#v:onChangeWith

```haskell
-- contrived example

data Action = Changed MisoString DOMRef | Process MisoString

update = \case
  Process selected -> io_ $ do
    consoleLog selected
  Changed newText domRef -> do
    io $ do 
       txt :: MisoString <- fromJSValUnchecked =<< domRef # "select" $ ()
       pure (Process txt)
       
myView :: View m Action
myView = input_ [ onChangeWith Changed ]       
```

This will avoid the `getElementById` lookup, but only works if you're interested in calling `select()` in response to `<input>` value change events.

ts/miso.ts
```diff
@@ -5,6 +5,7 @@ import { hydrate, integrityCheck } from './miso/hydrate';
 import {
    version,
    callFocus,
+   callSelect,
```
dmjio: @Reijix also, `bun run js` should regenerate the JS, this should be committed to the repo as well.

Reijix:
Thanks for the feedback!
I've implemented all of it and have now added `setSelectionRange`, I decided to always set the optional parameter `selectionDirection` to the default `none`. This could of course also be exposed, however you prefer it.

dmjio:
The defaults are fine imo, we can generalize on an as-needed basis if need be. 

Thank you for your contribution 🍜 @Reijix

PR #1255, opened 2025-12-02
Set `componentId` on `VComp<T>`.

dmjio:
This change sets `componentId` on the virtual DOM (as opposed to the physical DOM).

- [x] Modifies `unmount` to take in the virtual DOM as argument
- [x] Moves logic to set `componentId` on `VComp`
- [ ] Add playwright test.


dmjio:
Subsumed by https://github.com/dmjio/miso/commit/af6d075dcf31e46c599068bc23936572dc8ea19f

PR #1263, opened 2025-12-05
Add `Property` data type

dmjio:
This should remove intermediate allocations of `Value` during `View m a` construction in `buildVTree`, for select `prop` combinators. Using the most generic `prop` combinator is now discouraging in favor of more specific `textProp`, `doubleProp`, `intProp`.

- [x] Add `Property` data type.
- [x] Define `integerProp` in terms of `doubleProp`.

There is no `ToJSVal` on `Integer`, users can marshall their `Integer` into a `Double`, or we can add a custom `BigInt` (non-IE compatible). Previously `integerProp` was marshalled into a `Scientific` value and then converted into a JS `float`.

Defining `integerProp` in terms of `doubleProp` is identical to this.

This should have a dramatic performance improvement on rendering large strings, since we now bypass calls to `textFromJSString`.


dmjio:
Make sure this is tested on the JS backend, w/ WASM it might be slower because it uses `Text` for everything since its using `jsaddle` command protocol and not raw FFI backend. Similar to #1264

PR #1264, opened 2025-12-05
Add efficient event parsing.

dmjio:
Currently, decoding events uses `aeson`. This works by converting the entire JS object structure into `Value`. This incurs significant allocation overhead. Events are high frequency and minimizing allocations is important.

This PR introduces an `Event` parser (`Parser a`) that bypasses the conversion into `Value` by performing parsing in `JSM` directly on the raw `JSVal` (`FFI.Event` object).

This works by replacing the `FromJSON` constraint in favor of the more efficient `FromJSVal` and parsing in an `ExceptT`, with two error cases.

We expose a `FromJSON`-like interface for end users. The code stays the same, but

- [x] Define and expose `withObject`, `(.:)`
- [x] Adds `SyntheticEvent` nomenclature.
- [x] Adds `executeDecoder` helper
- [x] Adds various `FromJSVal` instances.

~~This opens up the door for more aeson-combinators like `(.:?)`, `(.!=)`, etc.~~

- [x] `.:?`, `.!=` added


dmjio:
Make sure this is tested on the JS backend, w/ WASM it might be slower because it uses `Text` for everything since its using `jsaddle` command protocol and not raw FFI backend.

Issue #1257, opened 2025-12-03
Functions on DOMRef


Reijix:
`DOMRef` can currently only be modified/accessed by adding jsaddle to your project, does it make sense to reexport some functions on `DOMRef`?

I'm thinking of
- `(!)` - accessing properties
- `(<#)` - setting properties (note this clashes with `(<#)` from `Miso.Effect`
- `(#)` - calling a function
from `Language.Javascript.JSaddle.Object`
and
- `fromJSVal` - for casting
from `GHCJS.Marshal`

dmjio:
We could, but ideally users wouldn't even need to know `jsaddle` is being used. I personally don't like to re-export functions from other packages unless its `base` / `text`, those that don't change much. The `<#` is unfortunate, that was added before `miso` was using `jsaddle`, we used to rely only on the FFI explicitly, no `JSM` DSL.

We could export our own functions that wrap these, maybe give them friendlier names, like `callFunction`, `getProperty`, `setProperty`,  `marshall`, etc.

Reijix:
> We could, but ideally users wouldn't even need to know `jsaddle` is being used. I personally don't like to re-export functions from other packages unless its `base` / `text`, those that don't change much. The `<#` is unfortunate, that was added before `miso` was using `jsaddle`, we used to rely only on the FFI explicitly, no `JSM` DSL.
> 
> We could export our own functions that wrap these, maybe give them friendlier names, like `callFunction`, `getProperty`, `setProperty`, `marshall`, etc.

Right, I think re-export was the wrong word for me to use, I did mean adding wrappers for these functions :)
See #1266

PR #1266, opened 2025-12-05
Add functions on DOMRef

Reijix:
As suggested in #1257 I have now added
- `getProperty`
- `setProperty`
- `callFunction`
- `fromDOMRef`

I did not add tests for these functions, since I'm not sure where to put them.

### DIFF
--- a/src/Miso/FFI.hs
+++ b/src/Miso/FFI.hs
@@ -27,10 +27,9 @@ module Miso.FFI
   , select
   , setSelectionRange
   , alert
-  , setProperty
   , getProperty
   , callFunction
-  , fromJSVal
+  , castJSVal
     -- ** Styles
   , addStyle
   , addStyleSheet
--- a/src/Miso/FFI/Internal.hs
+++ b/src/Miso/FFI/Internal.hs
@@ -66,10 +66,9 @@ module Miso.FFI.Internal
    , diff
    , nextSibling
    , previousSibling
-   , setProperty
    , getProperty
    , callFunction
-   , fromJSVal
+   , castJSVal
    -- * Conversions
    , integralToJSString
    , realFloatToJSString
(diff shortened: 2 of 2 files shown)


dmjio:
Nice 🔥 

- Can we put these in `Miso.FFI.Internal`, and re-export out of `Miso.FFI` ?
- Instead of `DOMRef`, can we just expose it as `JSVal` (means we can drop `fromDOMRef`). Technically, all `DOMRef` are `JSVal`, but not all `JSVal` are `DOMRef`. So `callFunction` can operate on a vanilla JS object as well.
- Also, can we change the `setProperty` example, we wan't to discourage users from setting properties manually on the `DOM`, since the virtual DOM handles all of this.

Otherwise, LGTM 👌🏼

Reijix:
> Nice 🔥
> 
> * Can we put these in `Miso.FFI.Internal`, and re-export out of `Miso.FFI` ?
> * Instead of `DOMRef`, can we just expose it as `JSVal` (means we can drop `fromDOMRef`). Technically, all `DOMRef` are `JSVal`, but not all `JSVal` are `DOMRef`. So `callFunction` can operate on a vanilla JS object as well.
> * Also, can we change the `setProperty` example, we wan't to discourage users from setting properties manually on the `DOM`, since the virtual DOM handles all of this.
> 
> Otherwise, LGTM 👌🏼

Awesome, I've added most of it:

1. :heavy_check_mark: 
2. Yeah, makes sense! But `fromJSVal` still needs to be exported right? I've still added it as a wrapper now, or is it exported somewhere else?
3. Do you have a suggestion for a different example? I'm not sure what else `setProperty` might be used for, maybe it should be dropped altogether?

dmjio:
2. Maybe we could call it `castJSVal`, I'm not sure. There's really no way for us to get away from `jsaddle` because `JSM` is still required, albeit we only expose miso functions that operate in `JSM` to users.

3. Dropping it for now is fine imo. Don't want to encourage setting properties on the DOM outisde of miso's diff.

Reijix:
`castJSVal` sounds good! My goal is just to prevent having to import `jsaddle` in my project :)

PR #1292, opened 2025-12-19
Set parent reference on the first child of a `Component`.

dmjio:
This ensures that the parent reference is set appropriately when the virtual DOM is generated.

- [x] Introduces `setParentRef`, usage in `vcomp`, `vnode` and `vtext` cases.


dmjio:
Closing in favor of https://github.com/dmjio/miso/pull/1293

PR #1323, opened 2026-01-04
Add `inline` JS functionality.

dmjio:
The `inline` function takes an `Object` and code fragment as arguments and dynamically executes the fragment with the Object as the scope. The returned value is marshaled back into the Haskell heap via a `FromJSVal` instance.

This is made possible by the `Function` constructor.

No TH required, and no use of `eval()`.

```haskell
data Person = Person { name :: MisoString, age :: Int }
  deriving stock Generic
  deriving anyclass (ToJSVal, ToObject)
logNameGetAge :: Person -> IO Int
logNameGetAge = inline
  """
  console.log('age', age, 'name', name);
  return age;
  """
```

- [x] Adds `inline` function, and FFI wrapper code
- [x] Adds `default` signature for `ToObject`

```js
export function inline(code, context = {}) {
  const keys = Object.keys(context);
  const values = Object.values(context);
  const func = new Function(...keys, code);
  return func(...values);
}
```

### DIFF
--- a/tests/app/Main.hs
+++ b/tests/app/Main.hs
@@ -5,7 +5,9 @@
 {-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE DeriveAnyClass      #-}
 {-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE MultilineStrings    #-}
 {-# LANGUAGE TypeApplications    #-}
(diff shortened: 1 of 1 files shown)


dmjio:
This should always be preferred over `eval()` because 

1) It is secure, runs isolated and cannot alter other local variables
2) Is subject to performance optimization by JS engines

PR #1347, opened 2026-01-18
Make `events` global.

dmjio:
Now that event delegation is global and not `Component`-local, this PR lifts events out of `Component` and moves them to be an argument of `startApp` / `miso`.

This is a breaking UI change, but reflective better of the internal state of the application.

- [x] Drop `defaultEvents :: Events` from `Component`.
- [x] Hoist `Events` as as argument to `startApp` and `miso`.
- [x] Drop `undelegator`.
- [x] Guard delegator behind `isRoot`,
- [x] Thread global Events through `Component` recursion in `initialize`.
- [x] Drop corresponding JS `undelegator` related functions.

Similar to [react](https://react.dev) we delegate, but never undelegate, and all events are bound to `<body>`

### DIFF
--- a/tests/app/Main.hs
+++ b/tests/app/Main.hs
@@ -748,11 +748,11 @@ main = withJS $ do
 
     describe "Component tests" $ do
       it "Should mount one component" $ do
-        liftIO (startApp testComponent)
+        liftIO (startApp mempty testComponent)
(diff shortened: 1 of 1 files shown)


dmjio:
This increased performance on the 1000 component mount test by 10x (`1.0s` down to `100ms`)

PR #1350, opened 2026-01-22
feat: simple instance for At [a]

droshux:
Introduce an instance of `At` for `[a]`.
`take` and `drop` handle negatives and out of bounds indices safely already so no extra checks are needed here to avoid exceptions.


dmjio:
It looks good but can we use `splitAt` and add a test for get and set with the edge conditions in `miso/tests/app/Main.hs`

dmjio:
https://github.com/dmjio/miso/commit/dee2a5ce312b1547d34d8bcb0c9f2838bf59c122

PR #1362, opened 2026-01-25
Replace ghc-cleanup with simply-nix

ners:
This is an experiment to see how it compares.


dmjio:
Let's just try it out and if browsers are accumulating and `simply-nix` isn't purging them, then we can handle that issue if and when it arises.

PR #1366, opened 2026-01-27
Add `inline-js` `QuasiQuoter`.

dmjio:
The current implementation of `inline` does not use quasi quotation. This adds a QQ (`js`) for inline JavaScript.

```haskell
factorial :: Int -> IO Int
factorial n = [js|
  let x = 1;
  for (i = 1; i <= ${n}; i++) {
    x *= i;
  }
  return x;
|]
```

### DIFF
--- a/src/Miso/FFI/QQ.hs
+++ b/src/Miso/FFI/QQ.hs
@@ -65,13 +65,13 @@ import qualified Miso.FFI as FFI
 --
 js :: QuasiQuoter
 js = QuasiQuoter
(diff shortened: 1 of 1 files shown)


dmjio:
The last step here is a `Data JSVal`, `Data JSString` instances for `GHC.Wasm.Prim (JSVal, JSString)`. Seems to work for JS backend right now `120 <- factorial 5`

dmjio:
We don't actually need the `Data JSString` instance.

PR #1389, opened 2026-02-01
Specialize `io_` to `IO ()`

dmjio:
This is done to optimize usage of `inline-js`. When using the `js` QQ with `io_` we lose inference because `FromJSVal a` is too general.

If `io_` is specialized to `IO ()` (as opposed to `IO a`) we're able to correctly infer the `FromJSVal ()` return type and satisfy the type checker.

```haskell
update = \case
  Test -> io_ [js| return console.log('foo'); |]
```

^ Inference now succeeds.

- [x] `io_ :: IO a -> IO ()` -> `IO () -> IO ()`
- ~~[ ] Update `sample-app` to use `Miso.FFI.QQ (js)`~~

### DIFF
--- a/sample-app/Main.hs
+++ b/sample-app/Main.hs
@@ -1,13 +1,11 @@
 ----------------------------------------------------------------------------
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE QuasiQuotes       #-}
 {-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE CPP               #-}
 ----------------------------------------------------------------------------
 module Main where
 ----------------------------------------------------------------------------
 import           Miso
-import           Miso.FFI.QQ (js)
 import qualified Miso.Html as H
 import qualified Miso.Html.Property as P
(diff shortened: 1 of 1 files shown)


sample-app/Main.hs
```diff
@@ -54,8 +56,7 @@ updateModel :: Action -> Transition Model Action
 updateModel = \case
   AddOne        -> counter += 1
   SubtractOne   -> counter -= 1
-  SayHelloWorld -> io_ $ do
-    consoleLog "Hello World"
+  SayHelloWorld -> io_ [js| return console.log('hello world'); |]
```
ners:
```suggestion
  SayHelloWorld -> io_ [js| console.log('hello world'); |]
```

ners: I'm not sure if replacing `consoleLog ...` with `[js| console.log(...); |]` is a good idea here.

dmjio: good call yea, it also doesn't build w/ GHCJS86 in CI, so we can remove.

PR #1402, opened 2026-02-07
Adds end to end server-client testing to miso-tests

Zer0-:
This PR adds two tests:
    - TestApp: generates a random html page with a link, clicks the link and
      makes sure the app reacts to it (which implies hydration was successful)
    - TestBindingsApp: Creates a variable-length chain of components with
      variables bound to one another. The expectation is a change to
      the value of the first component will propagate all the way down each component
      and come back up to change a second value in the first component.

In order to test end-to-end playwright.ts was changed to be an http endpoint, that
way our haskell test server can control playwrite, which in turn points a browser at
the test server. 
    - Note: The changes to playwright.ts were almost entirely vibe-coded and then
      tested by hand. It could definitely be slimmed down but I lack the motivation
      atm.

There is also the server module TestServer that uses environment variables
to determine whether to link to the ghcjs or wasm payloads in the html, as well
as serving the initial html for each test. The tests rely on quickcheck's ioProperty,
for each iteration the actual http server is started in the background and the
test makes an http request to the playwright server to hit the server.

TODO:
- [ ]  add asserts to bindings test
- [ ] possibly rename the new binaries to something other than integration-* if that's more appropriate
- [x] comment out or delete serveFailed in TestServer (though it's useful)

I definitely expect a good amount of comments but I do think this way of testing is necessary.


nix/source.nix
```diff
@@ -42,6 +42,19 @@ in
...
+    rev = "00781d1920795b67e0476b67ed6840c388f29810";
+    sha256 = "sha256-dYPlwSbQ+QXvMeS5tonBVnT9zQGADtohmD/ZAiY/cXA=";
+  };
+
+  servant-miso-router = fetchFromGitHub {
+    owner = "haskell-miso";
+    repo = "servant-miso-router";
+    rev = "0c828e0ba30ee7a446ce8999288b32b7f6425dd1";
+    sha256 = "sha256-2Vkheb2iNDFWNAToO+r8rMY3OAA6LlUtgxiCWRm0wAY=";
+
```
dmjio, line 59: `}` 👀

PR #1404, opened 2026-02-09
Upstream dual-thread check from `miso-lynx`.

dmjio:
`miso-lynx` executes a single application on two JS interpreters (MTS / BTS). This PR introduces conditional checks to detect which interpreter is currently executing.

The main thread logic primarily runs in JS, so this PR currently exits the Haskell main thread, this is subject to change.

This PR only applies to the new JS backend, not WASM.

- [ ] Drop logging
- [x] Introduces `bts` function
- [x] Uses `delegator` from `eventContext`


dmjio:
Subsumed by https://github.com/dmjio/miso/pull/1556

PR #1405, opened 2026-02-09
feat: implement To/From JSON for GHC.Natural

droshux:
Added instances of `ToJSON` and `FromJSON` for `GHC.Natural`. 
Unmarshalling a negative number into a Natural will fail.

I've added tests but they're possibly in the wrong
place (or not even nessecary) because I couldn't see any other tests for JSON parsing.

### DIFF
--- a/tests/app/Main.hs
+++ b/tests/app/Main.hs
@@ -829,9 +829,9 @@ main = withJS $ do
       it "Should marshal a Natural" $ do
         JSON.fromJSON (JSON.toJSON (99 :: Natural)) `shouldBe` (JSON.Success (99 :: Natural))
         JSON.fromJSON (JSON.toJSON (0 :: Natural)) `shouldBe` (JSON.Success (0 :: Natural))
-        (JSON.fromJSON (JSON.Number $ -99.00) :: JSON.Result Natural) `shouldBe` (JSON.Error "cannot parse negative number as Natural")
-        (JSON.fromJSON (JSON.Number $ (0/0)::Natural) :: JSON.Result Natural) `shouldBe` (JSON.Error "cannot parse NaN as Natural")
-        (JSON.fromJSON (JSON.Number $ 15.24) :: JSON.Result Natural) `shouldBe` (JSON.Success 15)
+        (JSON.fromJSON (JSON.Number $ -99.00) :: JSON.Result Natural) `shouldBe` JSON.Error "cannot parse negative number as Natural"
+        ((JSON.fromJSON (JSON.Number $ 0/0)) :: JSON.Result Natural) `shouldBe` JSON.Error "cannot parse NaN as Natural"
+        (JSON.fromJSON (JSON.Number $ 15.24) :: JSON.Result Natural) `shouldBe` JSON.Success 15
       it "Should marshal a MisoString" $ do
         (`shouldBe` Just ("foo" :: MisoString)) =<< liftIO (fromJSVal =<< toJSVal ("foo" :: MisoString))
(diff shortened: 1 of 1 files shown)


src/Miso/JSON.hs
```diff
@@ -332,6 +336,11 @@ instance FromJSON Float where
 instance FromJSON Integer where
   parseJSON = withNumber "Integer" (pure . round)
 ----------------------------------------------------------------------------
+instance FromJSON Natural where
+  parseJSON = withNumber "Natural" parseNumber
+    where parseNumber d | d < 0 = pfail "cannot parse negative number as Natural"
+                        | otherwise  = pure $ naturalFromInteger $ fromInteger $ round d 
+----------------------------------------------------------------------------
```
dmjio, line 344: Can we handle the `isNaN` case,

Otherwise, this could happen

```haskell
ghci> naturalFromInteger $ fromInteger $ round $ ((0/0) :: Double)
269653970229347386159395778618353710042696546841345985910145121736599013708251444699062715983611304031680170819807090036488184653221624933739271145959211186566651840137298227914453329401869141179179624428127508653257226023513694322210869665811240855745025766026879447359920868907719574457253034494436336205824
```

and

```haskell
instance FromJSON Natural where
  parseJSON = withNumber "Natural" parseNumber
    where parseNumber d | d < 0 = pfail "cannot parse negative number as Natural"
                        | isNaN d = pfail "isNaN"
                        | otherwise  = pure $ naturalFromInteger $ fromInteger $ round d 
```

droshux:
> LGTM, but `isNaN` handling (with a test) and I'm unsure if the `toJSON` is actually lossy.

The `toJSON` implementation for Natural is lossy, because the implementation for Integer is lossy.
`toJSON = toJSON . naturalToInteger`

tests/app/Main.hs
```diff
@@ -825,6 +826,12 @@ main = withJS $ do
         (`shouldBe` Just (99 :: Int)) =<< liftIO (fromJSVal =<< toJSVal (99 :: Int))
         (`shouldBe` Just (-99 :: Int)) =<< liftIO (fromJSVal =<< toJSVal (-99 :: Int))
         (`shouldBe` Just (0 :: Int)) =<< liftIO (fromJSVal =<< toJSVal (0 :: Int))
+      it "Should marshal a Natural" $ do
+        JSON.fromJSON (JSON.toJSON (99 :: Natural)) `shouldBe` (JSON.Success (99 :: Natural))
+        JSON.fromJSON (JSON.toJSON (0 :: Natural)) `shouldBe` (JSON.Success (0 :: Natural))
+        (JSON.fromJSON (JSON.Number $ -99.00) :: JSON.Result Natural) `shouldBe` (JSON.Error "cannot parse negative number as Natural")
+        (JSON.fromJSON (JSON.Number $ (0/0)::Natural) :: JSON.Result Natural) `shouldBe` (JSON.Error "cannot parse NaN as Natural")
```
dmjio: I think this might need to be `Double`

```haskell
 (JSON.fromJSON (JSON.Number $ (0/0)::Double) :: JSON.Result Natural) `shouldBe` (JSON.Error "cannot parse NaN as Natural")
```

droshux: Fixed in https://github.com/dmjio/miso/pull/1405/commits/20ce421ffb3f9aa08336702fb2b4fe4900a52b69.
The issue was a brackets precedence thing, HLS was playing up so I wasn't aware of the issue until I saw the action failed.

PR #1408, opened 2026-02-10
Drop pointer indirection

dmjio:
`data` -> `newtype`


dmjio:
```
 src/Miso/Types.hs:269:6: error:
    • A newtype constructor cannot have a context in its type
      SomeComponent :: forall parent model action.
                       Eq model =>
                       Component parent model action -> SomeComponent parent
    • In the definition of data constructor ‘SomeComponent’
      In the newtype declaration for ‘SomeComponent’
    |
269 |    = forall model action . Eq model
    |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
    ```

PR #1430, opened 2026-02-21
Update `FromJSVal` / `ToJSVal` default generic deriving.

dmjio:
This uses a Generics implementation inspired by `jsaddle`.


dmjio:
https://github.com/dmjio/miso/pull/1432

PR #1437, opened 2026-02-23
Address `uriSub` in JS backend.

dmjio:
The JS backend does not seem to work well with the `createSub` utility (`bracket` pattern). WASM backend `Sub a` work fine with `createSub`.

This PR addresses the issues in `Sub` for the History API specifically related to the JS backend

- [x] Define and use `raisePopState`
- [x] Drop `Waiter` usage in `History` `Subscription`.
- [x] Use an infinite `threadDelay` in `createSub`

### DIFF
--- a/src/Miso/Subscription/History.hs
+++ b/src/Miso/Subscription/History.hs
@@ -24,12 +24,8 @@ module Miso.Subscription.History
   , URI (..)
   ) where
 -----------------------------------------------------------------------------
-import           Control.Concurrent
 import           Control.Monad
-import           Control.Monad.IO.Class
-import           System.IO.Unsafe
 -----------------------------------------------------------------------------
-import           Miso.Concurrent
 import           Miso.DSL
 import qualified Miso.FFI.Internal as FFI
 import           Miso.String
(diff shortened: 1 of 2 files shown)


Reijix:
I think manually raising `popstate` is not needed in `replaceState` and `pushState`, it is even in disagreement with the behaviour specified in https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event#the_history_stack:
> Note that just calling history.pushState() or history.replaceState() won't trigger a popstate event.

The fix to `createSub` should be enough to restore functionality for the js backend.

dmjio:
In our case it is needed because `uriSub` should be responsive to *all* `URI` changes. Calling `pushState` will modify the history stack and change the address bar, but the user's `model` will not be affected. Piggybacking on `popstate` solves this problem (since it sinks the results of `getURI`), and it also removes the `MVar` usage (was using `Waiter`) that was also causing blocking behavior it seems (similar to `createSub`).

dmjio:
I've changed `createSub` to no longer block on an `MVar` but to have an infinite thread delay, `unmount` and `stopSub` will call `killThread` and the finalizer will be invoked (listener removed cleanly to avoid browser leaks). 

This should solve the problem for all `Sub` usage in the JS backend (I noticed `keySub`, etc. were also affected).

dmjio:
I tested this against the `miso-svg` example, but modified to use `startSub` / `stopSub`, for both JS and WASM backends. Along with the history API.

PR #1438, opened 2026-02-24
Fix binding <-- operator

Zer0-:
- fixes transitive upwards bindings
    - where before the value would ony be copied once to its immediate parent
- stops the component that triggered the depth first search from propagateChildren
- this lets chained bindings copy the changed value in one go

### DIFF
--- a/src/Miso/Runtime.hs
+++ b/src/Miso/Runtime.hs
@@ -326,7 +326,7 @@ propagate vcompId vcomps =
 -----------------------------------------------------------------------------
 -- | Create an empty DFS state
 dfs :: IntMap (ComponentState p m a) -> ComponentId -> DFS p m a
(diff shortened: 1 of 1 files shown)


Zer0-:
This was manually tested against TestBindingsApp, and miso-reactive, and serves as a preliminary patch to eventually merging [this pr](https://github.com/dmjio/miso/pull/1402)

Zer0-:
tested wasm and ghcjs backends with this

dmjio:
This requires a more granular approach that includes treating field names as nodes in the graph, as opposed to the entire data type. This will then allow us to properly handle cycles. Unfortunately that requires a `Generic` constraint on all user `model` (along with the `Eq` constraint) to access the underlying field name. So this is still being weighed, but for non-cyclical Binding it works.

PR #1439, opened 2026-02-25
Add cache invalidating query parameter `?v=${Date.now()}`.

dmjio:
This ensures `script` / `link` assets are fetched fresh (when using `reload` and `:r`) with WASM browser mode.

Adds a conditional `CacheBust` argument to `Src` / `Href`.


dmjio:
Related to https://gitlab.haskell.org/ghc/ghc/-/merge_requests/15574

PR #1466, opened 2026-03-20
hls: 2.10 -> 2.12

yaitskov:
Hi,

I tried dev env `nix develop .#hls` and noticed that emacs lsp-mode cannot work with hls-2.10.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -43,11 +43,11 @@
       },
       "locked": {
         "host": "gitlab.haskell.org",
-        "lastModified": 1772669130,
-        "narHash": "sha256-uomiPFFhHOzYU/2t3iAczlUOczDfJWYMdyqjOPKNnig=",
+        "lastModified": 1773434424,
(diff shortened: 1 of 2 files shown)


dmjio:
Hi @yaitskov 👋🏼 

Instead of bumping the entire pkg set its probably better if we add latest `hls` to our `source.nix`. Then we can `callCabal2nix` on it in our `ghc9122` pkg set, and reference it in the flake `#hls` dev env. Using `miso-from-html` as an example below.

add new `hls` to `source.nix` (can use `nix-prefetch-git` for this):
- https://github.com/dmjio/miso/blob/master/nix/source.nix#L39-L44

Add `hls` to `ghc9122` pkg set:
- https://github.com/dmjio/miso/blob/master/nix/haskell/packages/ghc/default.nix#L13

update `flake.nix`, adding here:
- https://github.com/dmjio/miso/blob/master/flake.nix#L87

dmjio:
@yaitskov, I expected this to break the js backend derivation, which would require recaching. The native compiler also was unaffected. This appears to have been a safe bump. I can populate `haskell-miso-cachix` cache with `hls`.

PR #1477, opened 2026-03-30
wrap CPP init into TH

yaitskov:


dmjio:
@yaitskov is the goal of this feature to bypass the boilerplate associated with the `foreign import` declaration under the WASM backend? Would you still like to see this implemented?

yaitskov:
@dmjio, the goal is to reduce reliance on CPP macros that user have to see. 
IMHO a hello world app requires to use CPP makes bad impression for newbies. 
They could think that if even hello world resort to CPP the library is highly experimental and might lose interest in  trying.

dmjio:
Yes I agree, okay well we can export this top level conditionally for WASM users to consume in their apps.

PR #1497, opened 2026-04-14
Drop `MisoString` argument from `textarea_`.

dmjio:
Instead of using the `MisoString` argument to set `value`, normalize `value_` property to set `text` argument when rendering HTML on the server.


dmjio:
This also introduces `optionalVoidAttrs`, this allows users to conditionally set `.value` on `textarea`.

Issue #1500, opened 2026-04-21
Bindings should override initModel values on mount


mtamc:
**Describe the bug**

It seems that currently, values bound using `Miso.Binding` are not synced on mount, instead using whatever's inside your `initModel`. They become synced the next time the component is rendered(?), or some other post-mount event.

We need to write code like the below as a workaround.

```hs
component_ :: parent -> (parent -> [Attribute Action]) -> Component parent Model Action
component_ parent parentToAttrs =
    (Miso.component (initModel parent parentToAttrs) updateModel viewModel)
        { Miso.bindings = [ ParentToChild parentToAttrs (#attrs .~) ]
        }

data Model
    = Model
    { myChildStuff :: MisoString
    , attrs :: [Attribute Action]
    }
    deriving (Eq, Show, Generic)
    
initModel :: parent -> (parent -> [Attribute Action]) -> Model
initModel parent parentToAttrs =
    Model { myChildStuff = "", attrs = parentToAttrs parent }

viewModel :: Model -> View Model Action
viewModel model =
    H.div_ (secretAttrs <> model.attrs) [H.text "Hi"]
```

**Expected behavior**
If a binding is defined, it should immediately override any initial value on mount, not wait until later. The code we should be able to write looks like:

```hs
component_ :: (parent -> [Attribute Action]) -> Component parent Model Action
component_ parentToAttrs =
    (Miso.component initModel updateModel viewModel)
        { Miso.bindings = [ ParentToChild parentToAttrs (#attrs .~) ]
        }

data Model
    = Model
    { myChildStuff :: MisoString
    , attrs :: [Attribute Action]
    }
    deriving (Eq, Show, Generic)

initModel :: Model
initModel =
    Model { myChildStuff = "", attrs = [] }

viewModel :: Model -> View Model Action
viewModel model =
    H.div_ (secretAttrs <> model.attrs) [H.text "Hi"]
```

**Screenshots**
N/A

**Desktop (please complete the following information):**
Browser-agnostic

**Smartphone (please complete the following information):**
Browser-agnostic

**Additional context**
Follows conversation on Matrix, search up "bindings"

PR #1501, opened 2026-04-21
Inherit parent `model` state on `Component` `mount`.

dmjio:
When a `Component` mounts, it should inherit its parent state (if `ParentToChild` binding is in use). Similarly, if `ChildToParent` is in use, the `parent` state should be overridden by the child state when mounting.

- [x] Adds `inheritParentBindings` and `inheritChildBindings` in `Runtime`

### DIFF
--- a/src/Miso/Binding.hs
+++ b/src/Miso/Binding.hs
@@ -32,11 +32,16 @@
 module Miso.Binding
   ( -- ** Types
     Binding (..)
+  , Precedence (..)
     -- ** Combinators
   , (<-->)
+  , (<<-->)
+  , (<-->>)
   , (<--)
   , (-->)
   , (<--->)
+  , (<<--->)
+  , (<--->>)
   , (<---)
   , (--->)
   ) where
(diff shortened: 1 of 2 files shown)


dmjio:
Related to #1500

mtamc:
I have tested this commit with a `ParentToChild` binding and it appears to resolve #1500. 
Untested for `ChildToParent`. Unsure if there was any bug to begin with for bidirectional bindings.

mtamc:
Update: This may not be fixed for bidirectional bindings. (Will confirm later if I get the time to make a minimal repro.)

dmjio:
For bidirectional it's not clear which state should take precedence, the parent or the child. 

I think we might only be able to support `ParentToChild` on `mount`

mtamc:
It would be great to be made into a parameter. Maybe it would simply look like `<-->'` (Miso lenses) and `<--->'` (VL lenses) to make the parent take precedence. `<-->` `<--->` would keep having the child take precedence.

dmjio:
I added `<<-->`, `<-->>` to determine if `Child` or `Parent` `Component` state should be inherited on `mount`.

mtamc:
I tested this locally and it works as expected. This would close #1500

PR #1508, opened 2026-05-04
Support overloaded labels for Miso.Lens

goosedb:
Adds an `IsLabel` instance for `Miso.Lens`, allowing the use of `#label` syntax as lenses when `OverloadedLabels` is enabled.

This is a small quality-of-life improvement. No strong feelings if this is deemed unnecessary or inappropriate for the project — happy to discuss or have this declined.

### DIFF
--- a/src/Miso/Lens/Generic.hs
+++ b/src/Miso/Lens/Generic.hs
@@ -11,7 +11,7 @@
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Miso.Lens.Generic where
+module Miso.Lens.Generic () where
 
 import Data.Kind (Constraint, Type)
 import GHC.Generics (C1, D1, Generic (..), K1 (..), M1 (..), Meta (..), Rec0, S1, U1, V1, (:*:) (..), (:+:) (..))
--- a/src/Miso/Lens/Generic.hs
+++ b/src/Miso/Lens/Generic.hs
@@ -17,8 +17,7 @@ import Data.Kind (Constraint, Type)
 import GHC.Generics (C1, D1, Generic (..), K1 (..), M1 (..), Meta (..), Rec0, S1, U1, V1, (:*:) (..), (:+:) (..))
 import GHC.OverloadedLabels (IsLabel (..))
 import GHC.Records (HasField (..))
-import GHC.TypeError (ErrorMessage (..), TypeError)
-import GHC.TypeLits (Symbol)
+import GHC.TypeLits (ErrorMessage (..), Symbol, TypeError)
 import Miso.Lens (Lens, lens)
 
 instance
@@ -28,7 +27,6 @@ instance
   fromLabel :: (HasField name s a, Generic s, GSet name a (Rep s)) => Lens s a
   fromLabel = lens (getField @name) (\s v -> to . gSet @name v . from $ s)
   {-# INLINE fromLabel #-}
-
 
 class GSet (name :: Symbol) typ f where
   gSet :: typ -> f x -> f x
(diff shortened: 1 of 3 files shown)


dmjio:
This is a very welcome addition. I'd also be interested in creating something akin to what `generic-lens` has w/ `field'` if possible, but using our `Lens` formulation in `Miso.Lens`.

```haskell
class HasField' (field :: Symbol) s a | s field -> a where 
  field' :: VL.Lens s s a a

-- >>> getField @"age" human 50 
-- getField :: forall f a s. HasField' f s a => s -> a 
-- getField = VL.view (field' @f) 
--
-- >>> setField @"age" 60 human 
-- Human {name = "Tunyasz", age = 60, address = "London", other = False} 
--
-- setField :: forall f s a. HasField' f s a => a -> s -> s 
-- setField = VL.set (field' @f)  
```

goosedb:
We also need to figure out: maybe it’s better to put the IsLabel instance right into the Miso.Lens module, instead of leaving it as an orphan in Miso.Lens.Generic?

dmjio:
I'm fine leaving it in its own module for now as an orphan. I don't anticipate any problems with orphan instances. It's also easier to gatekeep it behind a flag in the cabal file so we can disable it if need be. For example, we plan on supporting MicroHs regardless of [type family support](https://github.com/augustss/MicroHs/issues/367), so the ability to remove the module from compilation is valuable. Otherwise we have to CPP things and it can get messy w/ imports.

Do you mind adding a module header like the others w/ your name, etc. and formatting like the other modules.

dmjio:
Also, it seems your definition of `fromLabel` is what `field'` from `generic-lens` accomplishes. Can we factor it out into its own top-level declaration (e.g.`field`) and then export that?

```haskell
field :: (HasField name s a, Generic s, GSet name a (Rep s)) => Lens s a
field = lens (getField @name) (\s v -> to . gSet @name v . from $ s)
```

goosedb:
Ok! I’ll get it done tomorrow or the day after.

goosedb:
Please review the module interface

src/Miso/Lens/Generic.hs
```diff
@@ -0,0 +1,103 @@
...
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
+{-# LANGUAGE FunctionalDependencies #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Miso.Lens.Generic
+-- Copyright   :  (C) 2016-2026 David M. Johnson
+-- License     :  BSD3-style (see the file LICENSE)
+-- Maintainer  :  David M. Johnson <code@dmj.io>
```
dmjio, line 20: Feel free to add yourself for attribution sake (if you'd like).

src/Miso/Lens/Generic.hs
```diff
@@ -0,0 +1,103 @@
...
+  (HasField name s a, TotalityCheck name s a (GetFieldType name (Rep s)), GSet name a (Rep s), Generic s) => 
+  HasLens name s a where
+  getLens = lens (getField @name) (\s v -> to . gSet @name v . from $ s)
+  {-# INLINE getLens #-}
+
+instance HasLens name s a => IsLabel name (Lens s a)
+  where fromLabel = getLens @name
+
+{-# INLINE field #-}
+field :: forall name s a. HasLens name s a => Lens s a
```
dmjio, line 49: This is perfect, would you care to add a test for this in `tests/app/Main.hs`, around [here](https://github.com/dmjio/miso/blob/master/tests/app/Main.hs#L129). We have a sample `Person` data type w/ `Generic` instance that could be used to get / set the `name` / `age` field, etc.

src/Miso/Lens/Generic.hs
```diff
@@ -0,0 +1,103 @@
...
+import Data.Kind (Constraint, Type)
+import GHC.Generics (C1, D1, Generic (..), K1 (..), M1 (..), Meta (..), Rec0, S1, (:*:) (..), (:+:) (..))
+import GHC.OverloadedLabels (IsLabel (..))
+import GHC.Records (HasField (..))
+import GHC.TypeLits (ErrorMessage (..), Symbol, TypeError)
+-----------------------------------------------------------------------------
+import Miso.Lens (Lens, lens)
+-----------------------------------------------------------------------------
+
+class Generic s => HasLens (name :: Symbol) s a | name s -> a where
```
dmjio, line 36: `HasLens` class looks good, fine with me.

dmjio (changes requested):
The `HasLens` class is fine with me.

src/Miso/Lens/Generic.hs
dmjio: I added a test

goosedb:
Oh, I'm late :( 
I've also tried to write tests for compile errors too (with `-fdefer-type-errors`), but since a type error is kind of "phantom" it doesn't work (setting missing field is just `id` by implementation)

PR #1509, opened 2026-05-05
Add `keyed`

dmjio:
Allows the appending of `Key` to any `View`

```haskell
keyed "key" ("some text" :: View model action)
keyed "key" $ div_ [ id_ "container" ] [ "content" ]
keyed "key" (mount_ calendarComponent)
```


dmjio:
cc @ners

PR #1516, opened 2026-05-18
Add React-style props passing between `Component`

dmjio:
## Motivation

Components often need data from their parent — a selected id, a theme, a page limit — that tracks the parent's model as it changes. Props provide a direct render-time channel for this: the parent passes a value via `mountProps` and the child receives an action whenever that value changes.

This is modelled after React's props system
(https://react.dev/learn/passing-props-to-a-component), where a parent passes data to a child at render time and the child re-renders when those values change. The key difference is that Miso components own their own model and update loop, so props arrive as a dispatched action via `useProps` rather than being injected directly into the view.

## Feature overview

A parent component can now pass any JSON-serialisable value to a child using `mountProps`. The child opts in by defining `useProps` on its `Component` record. Each time the parent re-renders with changed props, the diff engine calls the child's `useProps` handler, which may dispatch an action to update the child's model.

Change detection is done in Haskell by comparing the new `Value` against `_componentLastProps` (stored purely in `ComponentState` — no extra `IORef`). The child's `useProps` is only invoked when the props value actually changes, so re-renders are not triggered spuriously.

## Example

```haskell
-- Parent view
view :: Model -> View Action
view m =
  mountProps_ "themePanel" (m ^. limit, m ^. theme) $
    (component 0 childUpdate childView)
       { useProps = checkProps PropsReceived BadProps }

-- Child action
data ChildAction = PropsReceived (Int, Theme) | BadProps MisoString | ...

childUpdate :: ChildAction -> Effect ChildModel ChildAction
childUpdate (PropsReceived (limit, theme)) = modify \m -> m { limit, theme }
childUpdate (BadProps err) = io_ (consoleError err)
```

## Summary of changes

- `Miso.Types`
  - New `Props = Maybe Value` type alias
  - New `useProps :: Value -> Maybe action` field on `Component` (defaults to `const Nothing`)
  - `VComp` constructor gains a `Props` slot
  - New `mountProps` combinator (keyed mount with props)
  - `mount_`, `(+>)`, `keyed` updated to pass `Nothing` for props
  - Haddock example on `mountProps` updated to use `checkProps`

- `Miso.Runtime`
  - `_componentLastProps :: Maybe Value` replaces the former `IORef (Maybe Value)` — pure field updated via `modifyComponent`
  - New `componentLastProps` lens
  - New `checkProps` helper (mirrors `checkMail`) for decoding props into actions
  - `buildVTree` installs a `sendProps` JS callback when props are present; the callback reads the component from the global map, compares, and dispatches only on change

- `Miso.hs`
  - Exports `mountProps` and `checkProps` under a new `-- ** Props` section

- `Miso.Html.Render` — SSR pattern updated for new `VComp` shape

- TypeScript / JS
  - `VComp` type gains optional `sendProps?: () => void`
  - `diff` calls `n.sendProps()` whenever a component node is reused (i.e. patched in-place rather than replaced), so the child receives fresh props on every parent re-render


dmjio:
https://gist.github.com/dmjio/4d228b0a91eb56cc59700b53f16df134 <-- prop drilling example

PR #1540, opened 2026-06-19
Store `schedulerThread` in `StablePtr`.

dmjio:
When using `live` and `reload` functions, the scheduler `ThreadId` needs to persist between `:r` invocations (via `StablePtr`). Otherwise, the CAF gets recreated and the previous scheduler thread (along with its closure) leaks into the heap (due to the `forever` call). This causes accumulated heap growth during a hot reload session. By storing the `ThreadId` in the `StablePtr` this allows us to kill the previous scheduler `ThreadId` cleanly.

Likewise the `ComponentState`, in order to be reified reliably, needs to be fetched by unique `Key` (not `ComponentId`, which is not stable between user changes to application state). Onus falls on the user to `Key` their `Component` uniquely during a hot-reload session (with the exception of the top level `ComponentId` which is always 1).

This means `live` users should always `Key` `Component` uniquely. The only `Component` that is allowed to be left unkeyed during hot-reload (w/ `live`) is the top-level root `Component`, of id `1`.

- [x] Add `ThreadId` to `StablePtr` in `live` / `reload` (along with `components :: IORef (CompenentState parent props model action)`)
- [x] Don't `clearPage` during `live`, just `clearBody`. Scripts / styles only need to be loaded once.
- [x] Drop `killThread =<< readIORef schedulerThread`
- [x] Add `_componentKey` to `ComponentState`.

### DIFF
--- a/src/Miso/Reload.hs
+++ b/src/Miso/Reload.hs
@@ -105,7 +105,7 @@ reload events vcomp = do
   when (exists == 1) $ do
     (_, oldSchedulerRef) <- deRefStablePtr =<< x_get
     killThread =<< readIORef oldSchedulerRef
-  x_clear
+    x_clear
   clearPage
   void (initComponent events Draw False vcomp)
   x_store =<< newStablePtr (components, schedulerThread)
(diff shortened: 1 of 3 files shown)


dmjio:
Consider constraining `model` on `Generic`

PR #1556, opened 2026-07-06
feat: add Lynx thread detection (`onBTS` / `onMTS`)

dmjio:
### 🧵🧵 dual-thread support

`miso` now detects at runtime whether it is executing on a Lynx Background Thread (BTS), Main Thread (MTS), or a plain web build, and skips browser-only side-effects (script injection, DOM cleanup, `mountComponent`, `scheduler`) where they do not apply.

- Add `onBTS()` / `onMTS()` TypeScript helpers in `ts/miso/util.ts` using `typeof __BACKGROUND__` / `typeof __MAIN_THREAD__` guards so they safely return `false` in web builds
- Add `ts/globals.d.ts` declaring the rspeedy compile-time globals `__BACKGROUND__` and `__MAIN_THREAD__` as `boolean | undefined`
- Export `onBTS` / `onMTS` from `ts/miso.ts` and `ts/index.ts`; register both on `globalThis['miso']`
- Add `ts/globals.d.ts` to `tsconfig.lib.json` `files` list
- Add `ts/spec/onBTS.spec.ts` with 6 tests (3 per function) covering the undefined, true, and false cases
- Rebuild `js/miso.js` and `js/miso.prod.js` with the new functions
- Add `onBTS :: IO Bool` and `onMTS :: IO Bool` to `Miso.FFI.Internal`; fix `delegator` to call via `getEventContext` instead of `jsg "miso"`
- In `Miso.Runtime.initialize`: skip `renderScripts`/`renderStyles` on Lynx; only call `FFI.mountComponent` on BTS
- In `Miso.Runtime.initComponent`: skip `cleanup` on Lynx; gate scheduler fork to BTS or web; load `miso.js` once via a `loadedJS` `IORef` to avoid double-evaluation


src/Miso/Runtime.hs
```diff
@@ -250,7 +255,7 @@ initialize events _componentParentId hydrate isRoot initialProps maybeKey comp@C
   initSubs subs _componentSubThreads _componentSink
   initialDraw _componentModel events hydrate isRoot comp vcomponent
   forM_ mount _componentSink
-  FFI.mountComponent _componentId =<< toObject jsNull
+  when bts (FFI.mountComponent _componentId =<< toObject jsNull)
```
dmjio, line 264: We need to `ToJSON model` constrain this (`renderApp`) eventually ... but it will probably require a cabal flag and some CPP.

PR #1560, opened 2026-07-14
Add await to miso DSL

Zer0-:
- Tested WASM and JS backends (but not legacy ghcjs)


dmjio:
I think legacy is picking it up
```
✓ Miso.DSL `await` tests > Successful Promise resolution should result in a value [0.600 ms]
✓ Miso.DSL `await` tests > A rejected Promise will throw an error we can catch [1.800 ms]
```

it LGTM

PR #1562, opened 2026-07-17
feat: upstream `miso-lynx` into `miso` as `Miso.Native` (dual-thread)

dmjio:
# 🍜 🐈 Dual-thread upstream 

This PR merges the separate `miso-lynx` project (https://github.com/haskell-miso/miso-lynx/pull/70) into `miso` proper, exposing native (Lynx) rendering under the `Miso.Native.*` namespace and reworking the runtime to conditionally support Lynx's dual-thread (main thread / background thread) architecture. 

There are now 4 protocols for communication, 2 in the Haskell layer, 2 in JS.

  - `COMPONENT` (`BTS` -> `MTS`), `model` hydration, `mount` / `unmount`
  - `EFFECTS` (`BTS` <-> `MTS`),

  - `PATCHES` (`BTS` -> `MTS`)
  - `EVENTS` (`MTS` -> `BTS`, `MTS` -> `MTS`)

`StaticPtr` is used to facilitate distributed computing across both interpreters via `WebWorker`. This means there is some additional `Component` boilerplate required now, specifically `-XStaticPointers`.

We also now require a `ToJSON` / `FromJSON` for both `model` / `action` parameters (serialization is required for cross-thread communication).

Example usage of dual thread operations below

```haskell
{-# LANGUAGE StaticPointers #-} 
import Miso
import Miso.JSON
import GHC.Generics
data Action
  = Ping
  | Pong
  deriving stock (Generic, Show, Eq)
  deriving anyclass (ToJSON, FromJSON)
main :: IO ()
main = native nativeEvents (static (mount_ counterComponent)) -- dmj: note the use of 'static' here
counterComponent :: App Int Action
counterComponent = (component 0 update_ view_)
  { mount = if mts then Just Ping else if bts then Just Pong else Nothing
  }
update_
  :: Action
  -> Effect context props Int Action
update_ = \case
  Ping ->
    runOnBG $ \sink -> do
      threadDelay (secs 1)
      consoleLog "ping"
      sink Pong
  Pong ->
    runOnMain $ \sink -> do
      threadDelay (secs 1)
      consoleLog "pong"
      sink Ping
```

# ✨ Highlights

* Native element tree — `Miso.Native` plus per-widget modules (`View`, `Text`, `Image`, `List`, `ScrollView`, `Frame`) each split into Element / Property / Event / Method, mirroring Lynx's component set. Adds `sample-app-native/` and native TS runtime (`ts/miso-native.ts`, `ts/miso/native/{bts,mts}`). Native flag `-fnative` also present to choose correct JS.

* `StaticPtr`-based `Component` identity — `VComp` now carries a `StaticPtr SomeComponent` instead of an inline component. The static key (`staticKey`/`deRefStaticPtr`/`unsafeLookupStaticPtr`) lets the main thread (MTS) locate a `Component` that is present on the background thread (BTS) (and vice versa) across the thread boundary, driving `mount` / `hydrate` / `unmount` over the `"Miso.components"` channel, and bidirectional effects over the `Miso.effects` channel.

* Removed the `parent` type parameter — `Component parent props model action` becomes `Component props model action`; `SomeComponent` and the `App` alias (`Component () model action`) drop `parent` accordingly. This removes the old parent/child `Binding` machinery (`Miso.Binding` deleted) in favor of the `props` / `mail` mechanisms. This also removes the 'model' from 'View model action', since `StaticPtr` necessitates monomorphic types.

* `MAIN_THREAD` / `BACKGROUND_THREAD` — the runtime detects which Lynx thread it is on (`__MAIN_THREAD__` / `__BACKGROUND__`). `Effects` carry an optional target `Thread` (`data Thread = MTS | BTS`); a `Schedule (Maybe Thread) ...` whose target is the opposite thread is not run locally but serialized and forwarded via `postEffect` so it executes same-thread on the other side. Adds `runOnMain` and the `mts`/`bts`/`web` boolean thread-detection helpers.

* Supporting changes to Runtime, Types, Effect, Reload, and the TS context/patch layers to route sinks, schedules, and hydration across the dual-thread boundary. `Nix`/`justfile`/`lynx.config` wiring for the native build pipeline.

* Add `sample-app-native` w/ canonical counter example

## Outstanding items

  - [ ] Add native polyfills for `WebSocket` / `Fetch` / `SSE` in Swift / Kotlin
  - [x] Add new / additional Lynx view types per https://lynxjs.org
  - [x] Add native FFI helpers for invoking transitions on MTS
  - [x] Haddocking on `Miso.Native`.
  - [x] Main thread events
  
 ## Optional "nice-to-haves"
 
  - [ ] `Miso.JSON.Diff`, only transfer JSON patches from BTS to MTS


js/miso.js
```diff
@@ -51,8 +51,8 @@ function fetchCore(url, method, body, requestHeaders, successful, errorful, resp
       for (const [key, value] of response.headers) {
         headers[key] = value;
       }
-      if (!response.ok) {
-        throw new Error(response.statusText);
+      if (response.status < 200 || response.status >= 300) {
+        throw new Error(response.statusText || "HTTP " + response.status);
```
dmjio, line 55: This is required because `lynx.fetch` does not implement `.ok`.

js/miso.js
```diff
@@ -491,9 +491,32 @@ function diffAttrs(c, n, context) {
   diffProps(c ? c.props : {}, n.props, n.domRef, n.ns === "svg", context);
   diffClass(c ? c.classList : null, n.classList, n.domRef, context);
   diffCss(c ? c.css : {}, n.css, n.domRef, context);
+  diffEvents(c, n, context);
```
dmjio, line 494: Only required for main thread event handlers, o/w all events by default arise on the background thread

js/miso.js
```diff
@@ -491,9 +491,32 @@ function diffAttrs(c, n, context) {
...
   diffChildren(c ? c.children : [], n.children, n.domRef, context);
   drawCanvas(n);
 }
+function diffEvents(c, n, context) {
+  if (!onBTS() && !onMTS())
+    return;
+  if (c === null && n.directEvents) {
+    for (const name of n.directEvents) {
+      context.addEvent(n.domRef, name, { capture: false, direct: true });
```
dmjio, line 503: Some element events require `__AddEvent` directly applied to their `node` (delegation won't work). This is made explicit now in `VNode`, there is a list for directly bound events.

js/miso.js
```diff
@@ -524,10 +547,11 @@ function diffClass(c, n, domRef, context) {
 }
 function diffProps(cProps, nProps, node, isSvg, context) {
   var newProp;
+  const native = onBTS() || onMTS();
```
dmjio, line 550: native only works w/ diffing attributes.

js/miso.js
```diff
@@ -601,6 +625,22 @@ function diffChildren(cs, ns, parent, context, endAnchor = null) {
     }
   }
 }
+function eventEntries(c, capture) {
```
dmjio, line 628: helpers for `diffEvents`

js/miso.js
```diff
@@ -1078,17 +1118,6 @@ var hydrationContext = {
...
-  mountComponent: function(componentId, model) {
-    return;
-  },
-  unmountComponent: function(componentId) {
-    return;
-  },
-  modelHydration: function(componentId, model) {
-    return;
-  }
-};
```
dmjio, line 1091: Dropped in favor of `COMPONENT` hydration protocol

sample-app-native/ios/App/NativeLocalStorageModule.m
```diff
@@ -0,0 +1,48 @@
+#import "NativeLocalStorageModule.h"
```
dmjio: example of native module polyfill for `localStorage`

src/Miso/Runtime.hs
```diff
@@ -246,11 +302,17 @@ initialize events _componentParentId hydrate isRoot initialProps maybeKey comp@C
...
-
   initSubs subs _componentSubThreads _componentSink
   initialDraw initializedModel events hydrate isRoot comp vcomponent
   forM_ mount _componentSink
-  when bts (FFI.mountComponent _componentId =<< toObject jsNull)
+#ifdef NATIVE
+  -- Ship the child's initial @props@ so the MTS can rebuild the mirror
+  -- component by applying the 'Props' constructor recovered from the
+  -- 'StaticKey'. The no-props case serializes @()@ (JSON @null@).
+  when (bts && not isRoot) $
```
dmjio: `not isRoot` because top-level component is mounted on both MTS and BTS on initial draw

js/miso.js
dmjio: Also, this is a no-op w/ web backend

Huxpro:
Hi @dmjio — I'd like to help upstream this in a reviewable stack.

Our team has been building across the Lynx frontend ecosystem (ReactLynx in lynx-stack, Octane Lynx, and Vue Lynx), and we've now exercised the miso native dual-thread path on real Lynx Sandbox devices rather than only reading the code. The core direction looks sound, especially preserving synchronous MTS first paint and app-specific `onMain` semantics.

I propose keeping three decisions independent:

1. **Foundation:** generated-bundle provenance/integrity, Nix source isolation, and a dual-thread conformance fixture.
2. **IFR correctness:** slot-normalized initial-frame reconciliation and node-id adoption, with a compact production digest and a safe BTS-authoritative full-tree repaint on mismatch.
3. **MTS architecture RFC:** measured full-MTS cost vs thin/B+/generated-slice modes. This would remain an issue/RFC so accepting IFR does not imply an MTS architecture decision.

For review topology, do you prefer these draft PRs to target the current `dual-thread` head from #1562, or target `master` and be rebased/retargeted with #1562? Unless you prefer the latter, I'll use `dual-thread` so each diff is directly reviewable against the code it extends.

I'll include the Sandbox forced-mismatch, delayed-handshake, and DevTool evidence in the IFR PR body; the local validation-only branches will not be published.

dmjio:
Hey @Huxpro,

A reviewable stack is very welcome. Direct PRs to this branch (`dual-thread`) would be preferable.

Huxpro:
Thanks — I used `dual-thread` directly as requested. The reviewable drafts and separate architecture discussion are now:

- #1577 — foundation: bundle provenance/integrity, Nix source isolation, and the core dual-thread conformance fixture
- #1578 — IFR: compact production digest, node-ID adoption, event/patch gates, and BTS-authoritative repaint on mismatch
- #1579 — MTS architecture RFC: A/B/B+/C decision, StaticPtr and offline-manifest experiments, and gradual rollout gates

The IFR PR body includes the exact Lynx Sandbox evidence. The local forced-mismatch and delayed-handshake probe branches were not published.

Both implementation PRs intentionally target `dual-thread`; #1578 lists `3fd4a91a..f80a078f` as its focused review range so foundation can be reviewed/landed first.

dmjio:
@Huxpro, 

#1577 and #1578 initially look good to me, but #1578 is larger than anticipated. Given this branch (`dual-thread`) is already very large in scope, I think it would be best now to merge it standalone and point #1577, #1578 to `master` (post merge). 

Then we can make a `1.13` release.  Also, have gone with option `A` on  #1579

Huxpro:
Thanks — done as requested.

- #1577 and #1578 now both target the merged `master@e92e444d`.
- Their histories are independent: #1578 no longer includes #1577, so its focused range is `e92e444d..7e62243c`.
- Both PR bodies have been refreshed and each PR has a new review request comment.
- #1579 now records Option A as the maintainer-selected primary architecture; B/B+/C remain experimental.
- Neither implementation PR currently has review threads requiring a code change or resolution.

Thanks for landing the dual-thread foundation and for the clear Option A rationale.

PR #1563, opened 2026-07-19
Add `React` `Context`. Parameterizes `Component` on `context`.

dmjio:
Per [React Context](https://react.dev/learn/passing-data-deeply-with-context), this adds a `context` object & type parameter, set at top-level `Component` mount time, available to all `Component` in `view`, and optionally updated in `Effect` via `modifyContext` / `putContext`.

Along with a `useContext :: Bool` for deciding if said `Component` should redraw in response to global `context` change (defaults to `False` for all child component and `True` for top-level `Component`).

This is motivated in part by adding an `i8ln` implementation that uses `context` to propagate language settings to all `Component`.

- [x] Updated `src/Miso.hs` to include `context` instructions.
- [x] Drop `Data.Binding` module and `bindings` field, and propagation logic.
- [x] Replace `parent` for `context`, replace `model` for `context` in `View`
- [x] Add `context` propagation phase in scheduler
- [x] Add / export a top-level `setContext` for use w/ SSR
- [x] Adds `startAppWithContext`, defaulting `startApp` to `()` ~ `props`
- [x] Adds `modifyContext` and `putContext`
- [x] Parameterize `ComponentInfo` w/ `context`
- [x] Parameterize `View` on `context`
- [x] Add `reloadWithContext` / `liveWithContext`
- [x] Drop `ROOT` (unnecessary now that `parent` is gone)
- [x] Parameterize `SomeComponent` on `context`
- [x] Drop `isDirty` and `_componentIsDirty`

`context` should always monomorphize given that `startApp` always infers `context ~ ()`. So this should work with `-XStaticPointers` and the #1562 feature.


dmjio:
Example below

```haskell
{-# LANGUAGE LambdaCase        #-}
{-# LANGUAGE CPP               #-}
module Main where
import           Miso
import qualified Miso.CSS as CSS
import           Miso.CSS.Color (Color(RGB))
import qualified Miso.Html as H
import           Miso.Lens
import           Miso.Reload (liveWithContext)
-- | The app-global React-style @context@.
--
-- Unlike a 'Component' @model@ (which is local to a single component) or @props@
-- (passed parent-to-child), the @context@ is a single value shared by __every__
-- component in the tree.
data Theme
  = Light
  | Dark
  deriving (Show, Eq)
toggleTheme :: Theme -> Theme
toggleTheme Light = Dark
toggleTheme Dark  = Light
-- | Foreground / background colors for a 'Theme'.
themeColors :: Theme -> (Color, Color)
themeColors Light = (RGB 0 0 0, RGB 255 255 255)
themeColors Dark  = (RGB 255 255 255, RGB 0 0 0)
-- | Top-level component model state
data Model
  = Model
  { _counter :: Int
  } deriving (Show, Eq)
counter :: Lens Model Int
counter = lens _counter $ \record field -> record { _counter = field }
-- | Top-level component actions
data Action
  = AddOne
  | SubtractOne
  deriving (Show, Eq)
-- | Entry point for a miso application.
--
-- We seed the app-global @context@ with the initial 'Theme'. Note that @live@
-- and @startApp@ fix the context to @()@, so a non-trivial context uses the
-- @WithContext@ variants ('liveWithContext' / 'startAppWithContext').
main :: IO ()
#ifdef INTERACTIVE
main = liveWithContext defaultEvents Light app
#else
main = startAppWithContext defaultEvents Light app
#endif
-- | WASM export, required when compiling w/ the WASM backend.
#ifdef WASM
#ifndef INTERACTIVE
foreign export javascript "hs_start" main :: IO ()
#endif
#endif
-- | Top-level component.
--
-- Note it does __not__ set @useContext = True@: the root only needs to change
-- the context, and it does not have to re-render when the context changes. Only
-- the mounted 'themeViewer' sub-component opts into context-driven re-renders.
app :: Component Theme () Model Action
app = component emptyModel updateModel viewModel
-- | Empty application state
emptyModel :: Model
emptyModel = Model 0
-- | Updates the top-level model
updateModel :: Action -> Effect Theme props Model Action
updateModel = \case
  AddOne      -> counter += 1
  SubtractOne -> counter -= 1
-- | Top-level view. Renders a counter and mounts the 'themeViewer'
-- sub-component, which is the piece that reads and reacts to the @context@.
viewModel :: Theme -> props -> Model -> View Theme Action
viewModel _theme _ x =
  H.div_
    []
    [ H.button_ [ H.onClick AddOne ] [ text "+" ]
    , text $ ms (x ^. counter)
    , H.button_ [ H.onClick SubtractOne ] [ text "-" ]
    , H.br_ []
    , "theme-viewer" +> themeViewer
    ]
-- | Sub-component actions
data ChildAction
  = SetTheme Theme
  deriving (Show, Eq)
-- | A mounted sub-component that reads the app-global @context@.
--
-- It sets @useContext = True@ so the runtime re-renders it whenever the context
-- changes — even though its own @model@ and @props@ never change. Without this
-- flag the component would not re-render on context changes and would show a
-- stale 'Theme'.
themeViewer :: Component Theme () () ChildAction
themeViewer = (component () updateChild viewChild) { useContext = True }
-- | Toggling the theme mutates the app-global @context@ with 'putContext' and
-- logs the new value so you can watch the context change in the console.
updateChild :: ChildAction -> Effect Theme props () ChildAction
updateChild = \case
  SetTheme newTheme -> do
    putContext newTheme
    io_ (consoleLog ("context changed to: " <> ms (show newTheme)))
-- | The sub-component reads the current 'Theme' from its @context@ argument
-- (the first argument to every 'view') and styles itself accordingly.
viewChild :: Theme -> props -> () -> View Theme ChildAction
viewChild theme _ _ =
  H.div_
    [ CSS.style_
        [ CSS.color foreground
        , CSS.backgroundColor background
        ]
    ]
    [ text $ "sub-component sees theme: " <> ms (show theme)
    , H.br_ []
    , H.button_ [ H.onClick (SetTheme (toggleTheme theme)) ]
        [ text "Toggle theme" ]
    ]
  where
    (foreground, background) = themeColors theme
```

PR #1577, opened 2026-08-13
[native] isolate bundle provenance and add a conformance fixture

Huxpro:
## Summary

This is the reviewable foundation layer for Miso Native after #1562 landed on `master`. It intentionally contains no initial-frame reconciliation or MTS architecture policy.

- isolate Nix package source inputs so generated and untracked artifacts cannot silently enter builds
- generate the Lynx bundle and checksum from the exact source closure used by the build
- add a small native dual-thread conformance fixture that exercises the main/background split

## Review shape

- Base: merged `master` at `e92e444d` (#1562)
- This PR: three foundation/provenance commits, `e92e444d..f123259f`
- Sibling: #1578 — IFR reconciliation, now independently based on `master`
- RFC: #1579 — Option A remains primary; B/B+/C remain experimental

#1578 no longer includes this PR's commits, so the two implementation diffs can be reviewed and landed independently.

## Validation

- TypeScript library and spec typechecks
- Bun baseline: 391 tests, 8,220 assertions
- Nix `sample-app-native-conformance-bundle` build
- generated bundle provenance/checksum verification
- rebase onto merged `master`: clean three-commit replay and `git diff --check`

GitHub Actions are rerunning on the rebased head.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -131,6 +131,7 @@ jobs:
          nix-build -A sample-app-native-bundle
          mkdir -p bundle-artifact
          cp result/main.lynx.bundle bundle-artifact/main.lynx.bundle
+         cp result/main.lynx.bundle.sha256 bundle-artifact/main.lynx.bundle.sha256
 
      - name: Upload Lynx bundle
        uses: actions/upload-artifact@v4
@@ -268,6 +269,7 @@ jobs:
        run: |
          mkdir -p sample-app-native/android/app/src/main/assets
          cp bundle-artifact/main.lynx.bundle sample-app-native/android/app/src/main/assets/main.lynx.bundle
+         cp bundle-artifact/main.lynx.bundle.sha256 sample-app-native/android/app/src/main/assets/main.lynx.bundle.sha256
 
      - name: (Android) assemble debug APK
        working-directory: sample-app-native/android
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -137,7 +137,9 @@ jobs:
        uses: actions/upload-artifact@v4
        with:
          name: lynx-bundle
-         path: bundle-artifact/main.lynx.bundle
+         path: |
+           bundle-artifact/main.lynx.bundle
+           bundle-artifact/main.lynx.bundle.sha256
          if-no-files-found: error
 
   build-macos:
--- a/default.nix
+++ b/default.nix
@@ -31,6 +31,16 @@ rec {
       styles = ./sample-app-native/styles.css;
     };
 
+  # Core-elements-only dual-thread conformance fixture. It is a separate
+  # executable so gallery feature coverage remains intact.
+  sample-app-native-conformance-bundle =
+    pkgs.mkLynxBundle {
+      name = "sample-app-native-conformance-bundle";
+      jsDrv = pkgs.pkgsCross.ghcjs.haskell.packages.ghcNative.sample-app-native;
+      exeName = "app-native-conformance";
+      styles = ./sample-app-native/conformance.css;
+    };
+
   # Android host APK (Kotlin + Lynx SDK), built reproducibly via nixpkgs
   # androidenv + the gradle mitm-cache dep fetcher. Embeds the freshly built
   # sample-app-native-bundle in assets. See nix/android.nix.
--- a/nix/source.nix
+++ b/nix/source.nix
@@ -2,12 +2,23 @@
 with lib;
 with (builtins.fromJSON (builtins.readFile ../flake.lock));
 let
-  make-src-filter = src: with lib;
+  make-src-filter =
+    { src
+    , excludedNames ? []
+    }: with lib;
     cleanSourceWith {
       inherit src;
       filter =
-        name: type: let baseName = baseNameOf (toString name); in
-         ((type == "regular" && hasSuffix ".hs" baseName) ||
+        name: type:
+        let
+          baseName = baseNameOf (toString name);
+          excluded =
+            [ ".git" ".github" ".gradle" ".stack-work" "coverage"
+              "dist" "dist-newstyle" "node_modules" "result"
+            ] ++ excludedNames;
+        in
+         (!elem baseName excluded && (
+         (type == "regular" && hasSuffix ".hs" baseName) ||
          (hasSuffix ".yaml" baseName) ||
          (hasSuffix ".cabal" baseName) ||
          (hasSuffix ".css" baseName) ||
@@ -17,8 +28,7 @@ let
          (hasSuffix ".js" baseName) ||
          (baseName == "README.md") ||
          (baseName == "LICENSE") ||
-         (type == "directory" && baseName != "tests") ||
-         (type == "directory" && baseName != "dist"));
+         (type == "directory")));
     };
 
   # fetch from flake
@@ -31,10 +41,18 @@ let
 in
 {
   # local sources
-  miso             = make-src-filter ../.;
-  miso-tests       = make-src-filter ../tests;
-  sample-app       = make-src-filter ../sample-app;
-  sample-app-native = make-src-filter ../sample-app-native;
+  # Keep package sources independent: changing a sample or native host must not
+  # invalidate the Miso library derivation.
+  miso = make-src-filter {
+    src = ../.;
+    excludedNames = [ "sample-app" "sample-app-native" "tests" ];
+  };
+  miso-tests = make-src-filter { src = ../tests; };
+  sample-app = make-src-filter { src = ../sample-app; };
+  sample-app-native = make-src-filter {
+    src = ../sample-app-native;
+    excludedNames = [ "android" "ios" "build" "styles.css" ];
+  };
 
   # non-flakified sources
   miso-from-html = fetchFromGitHub {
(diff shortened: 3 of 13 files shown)


Huxpro:
@dmjio, this is now rebased onto the merged `master` as requested.

- base: `master@e92e444d`
- focused range: `e92e444d..f123259f`
- three foundation/provenance commits only
- #1578 is now an independent sibling on `master`, not a dependent diff

There are currently no review threads on this PR. The earlier local Nix/provenance validation is recorded in the body, the replay was conflict-free, and Actions are rerunning on the new head.

Could you take another review pass when convenient?

sample-app-native/Conformance.hs
```diff
@@ -0,0 +1,298 @@
...
+    , view_
+      [ id_ "dynamic-slot"
+      , CSS.style_
+        [ CSS.width "100%"
+        , CSS.height "52px"
+        , CSS.flexShrink 0
+        , CSS.marginBottom "10px"
+        ]
+      ]
+      (if visible then [dynamicChild] else [])
```
dmjio: nit: this could be refactored as `[ dynamicChild | visible ]`

sample-app-native/Conformance.hs
```diff
@@ -0,0 +1,298 @@
...
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE LambdaCase         #-}
+{-# LANGUAGE OverloadedStrings  #-}
+{-# LANGUAGE RecordWildCards    #-}
+{-# LANGUAGE StaticPointers     #-}
+-----------------------------------------------------------------------------
+-- | Deterministic dual-thread conformance fixture using only Lynx core
+-- @<view>@ and @<text>@ elements. Optional gallery behaviors must not be able
+-- to truncate the tree before IFR, BTS, MTS, churn, or lifecycle probes run.
+module Main where
```
dmjio, line 13: I ran the conformance fixture on a physical iPhone and in the simulator, have confirmed the behavior is correct.

sample-app-native/Conformance.hs
```diff
@@ -0,0 +1,298 @@
...
+
+mtsPaintSection :: View context Model Action
+mtsPaintSection =
+  view_
+    [ id_ "mts-paint-shell"
+    , CSS.style_ (panelStyle <> [CSS.backgroundColor (CSS.RGB 51 65 85)])
+    ]
+    [ view_
+      [ id_ "mts-paint"
+      , event (static (VE.onTapMainWith PaintMTS))
```
dmjio, line 193: 🤌

sample-app-native/Conformance.hs
```diff
@@ -0,0 +1,298 @@
...
+childSection :: Int -> View () Model Action
+childSection count =
+  view_
+    [ id_ "child-slot"
+    , CSS.style_
+      [ CSS.width "100%"
+      , CSS.flexShrink 0
+      ]
+    ]
+    [ vcomp (ChildProps count) (static (mountStaticWithProps childComponent)) ]
```
dmjio, line 293: 🤌

nix/source.nix
```diff
@@ -31,10 +41,18 @@ let
...
+  # invalidate the Miso library derivation.
+  miso = make-src-filter {
+    src = ../.;
+    excludedNames = [ "sample-app" "sample-app-native" "tests" ];
+  };
+  miso-tests = make-src-filter { src = ../tests; };
+  sample-app = make-src-filter { src = ../sample-app; };
+  sample-app-native = make-src-filter {
+    src = ../sample-app-native;
+    excludedNames = [ "android" "ios" "build" "styles.css" "conformance.css" ];
```
dmjio, line 54: Nice find. This is a big improvement for reducing closure size. Although having both a blacklist and a whitelist may be confusing. There is a new `lib.fileset` in nix we could potentially use in a follow-up PR that essentially accomplishes the same as this, yet without both black and whitelists (only whitelist).

Claude came up with this

```nix
makeSrc = { root, include }:
  lib.fileset.toSource {
    inherit root;
    fileset = lib.fileset.unions include;
  };

miso = makeSrc {
  root = ../.;
  include = [ ../src ../miso.cabal ../README.md ../LICENSE ];
};

miso-tests = makeSrc { root = ../.; include = [ ../tests ]; };
sample-app = makeSrc { root = ../.; include = [ ../sample-app ]; };
```

Otherwise, I'm fine keeping it as is (as you have it) and we adjust the old derivation in a follow-up PR.

default.nix
```diff
@@ -31,6 +31,16 @@ rec {
...
 
+  # Core-elements-only dual-thread conformance fixture. It is a separate
+  # executable so gallery feature coverage remains intact.
+  sample-app-native-conformance-bundle =
+    pkgs.mkLynxBundle {
+      name = "sample-app-native-conformance-bundle";
+      jsDrv = pkgs.pkgsCross.ghcjs.haskell.packages.ghcNative.sample-app-native;
+      exeName = "app-native-conformance";
+      styles = ./sample-app-native/conformance.css;
+    };
```
dmjio, line 42: this is great

.github/workflows/main.yml
```diff
@@ -268,6 +269,7 @@ jobs:
        run: |
          mkdir -p sample-app-native/android/app/src/main/assets
          cp bundle-artifact/main.lynx.bundle sample-app-native/android/app/src/main/assets/main.lynx.bundle
+         cp bundle-artifact/main.lynx.bundle.sha256 sample-app-native/android/app/src/main/assets/main.lynx.bundle.sha256
```
dmjio, line 274: I'm not totally convinced we need the `sha256sum` in CI since since the bundle is produced by nix, which guarantees reproducibility and the avoidance of tampering. Although, some users may not be using nix e2e to produce bundles / perform an ARM build. In which case tampering could occur I suppose. 

Regardless, I'm fine keeping it. But I think there is a missing step in CI which was causing failure when the PR was rebased against `dual-thread`. The fix should be below.

```diff
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ build-bundle: Upload Lynx bundle
      - name: Upload Lynx bundle
        uses: actions/upload-artifact@v4
        with:
          name: lynx-bundle
-         path: bundle-artifact/main.lynx.bundle
+         path: |
+           bundle-artifact/main.lynx.bundle
+           bundle-artifact/main.lynx.bundle.sha256
```

^ this would need to get applied to make the build green I believe.

dmjio (changes requested):
It LGTM 🤌 

I believe the only remaining change is to add `main.lynx.bundle.sha256` to the `main.yml` CI file to green the build 🟢 Otherwise the below is just opining and nits.

```diff
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ build-bundle: Upload Lynx bundle
      - name: Upload Lynx bundle
        uses: actions/upload-artifact@v4
        with:
          name: lynx-bundle
-         path: bundle-artifact/main.lynx.bundle
+         path: |
+           bundle-artifact/main.lynx.bundle
+           bundle-artifact/main.lynx.bundle.sha256
```

sample-app-native/Conformance.hs
dmjio: image

sample-app-native/Conformance.hs
Huxpro: Applied exactly as suggested: the optional child now uses `[ dynamicChild | visible ]` in `dd070ad8`.

default.nix
Huxpro: Thank you—glad this boundary is useful.

sample-app-native/Conformance.hs
Huxpro: Thank you!

nix/source.nix
Huxpro: Agreed—I'm keeping this PR's scoped source isolation unchanged and leaving a whitelist-only `lib.fileset` migration to a focused follow-up.

sample-app-native/Conformance.hs
Huxpro: Thank you for testing the fixture on both a physical iPhone and the simulator, and for attaching the screenshot. This adds valuable iOS runtime confirmation to the existing evidence.

sample-app-native/Conformance.hs
Huxpro: Thank you!

.github/workflows/main.yml
Huxpro: Applied the suggested CI fix exactly in `dd070ad8`: the `lynx-bundle` artifact now uploads both `main.lynx.bundle` and `main.lynx.bundle.sha256`, so the Android job can download the pair it already consumes. Local YAML parsing and producer → upload → consumer path checks pass; CI will provide the end-to-end confirmation.

The sidecar remains useful for direct Gradle/Android Studio and non-Nix/ARM bundle paths, while the Nix path retains its own reproducibility guarantee.

Huxpro:
@dmjio, all seven review threads are now addressed.

The two code changes are in `dd070ad8`:

- adopted `[ dynamicChild | visible ]` for the optional fixture child;
- uploaded both the bundle and checksum sidecar in the `lynx-bundle` artifact.

Local `git diff --check`, workflow YAML parsing, and producer → artifact → Android consumer path checks pass. This environment does not have Nix/GHC, so the refreshed Actions run will provide the end-to-end build confirmation. Thank you also for validating the fixture on a physical iPhone and simulator.

I've replied to and resolved each review thread. Could you take another review pass when convenient?

dmjio (approved):

PR #1578, opened 2026-08-13
[native] reconcile the dual-thread initial frame before patches

Huxpro:
## Summary

Adds explicit initial-frame reconciliation (IFR) directly on top of the merged dual-thread implementation in `master`, without committing upstream to a different MTS architecture.

- reconcile independently rendered MTS/BTS trees with a compact, versioned structural digest
- adopt BTS-authoritative node IDs before releasing buffered events and incremental patches
- retain the full manifest only for debug/NACK diagnostics
- recover a mismatch with a BTS-authoritative ordinary-patch repaint, then continue incrementally
- normalize `insertBefore` recording semantics and reset MTS node/context state before recovery
- keep the runtime timing hook test-only (`__misoTestBeforeInitialFrame`)

## Review shape

- Base: merged `master` at `e92e444d` (#1562)
- Focused review range: `e92e444d..7e62243c`
- Six IFR commits; no #1577 foundation commits are included
- Foundation/provenance remains independent in #1577
- MTS architecture decision is recorded separately in #1579: Option A remains primary

## Failure policy

IFR mismatch is not hydration. MTS paint is discarded; BTS repaints using the normal patch executor and then resumes incrementals. The forced mismatch was deliberate fault injection: normal MTS/BTS `VTree`, node ID and `StaticKey` parity remains an invariant. The open host-policy question is whether to add reload, fatal, callback or telemetry behavior around the bounded safe recovery.

## Validation

Fresh after rebase:

- TypeScript library and spec typechecks
- Bun 1.2.18: 397 tests, 8,241 assertions, 0 failures
- checked-in native JS regenerated from the rebased TypeScript source
- `git diff --check`

Existing Lynx Sandbox evidence from the pre-rebase product head `f80a078f`:

- compact 29-node/109-operation adoption and reload
- early event + delayed BTS handoff delivered exactly once
- forced structural mismatch, full BTS repaint, then live incremental updates
- gallery smoke test (325 nodes / 1,211 operations)
- BTS update, MTS-only style handler and hydrated-state `onMain` all passed; the two-thread error console was empty

The rebase resolved only upstream import/test integration and regenerated artifacts; the IFR runtime state-machine changes replayed intact.


Huxpro:
@dmjio, this is now independently rebased onto the merged `master` as requested; it no longer contains #1577's foundation commits.

- base: `master@e92e444d`
- focused range: `e92e444d..7e62243c`
- six IFR commits
- fresh rebase validation: both TypeScript configs and Bun 1.2.18 (`397` tests / `8,241` assertions) pass; native JS was regenerated

There are currently no review threads on this PR. The body now distinguishes the normal parity invariant from the deliberate mismatch fault injection and retains the existing Sandbox evidence.

Could you take another review pass when convenient?

dmjio:
I'm curious the exact scenario under which IFR failure can occur. Currently, miso has a handshake that happens on load between BTS / MTS, so the initial `VTree` creation won't occur until `READY` + `READY_ACK` messages have been received. This should ensure `nodeId` synchronicity. Also, due to the fact that `buildVTree` does not evaluate any effects, I'm unsure that this failure scenario can actually happen without tampering. The `VTree` should be built in lockstep on `BTS` / `MTS`. So a scenario where the BTS runs ahead of the MTS (during initial draw) to deliver patches seems unlikely (given also that BTS seems to load slower). It says in the PR that this is a "forced structural mismatch" which means there is an assumption this invariant can be violated from within the confines of the framework.

This assumes the user is calling the defacto method

```haskell
main :: IO ()
main = native nativeEvents (vcomp (static theirComponent))
```

If `IO` is introduced before the call to `native` then there could be a discrepancy between `BTS` / `MTS` `READY` / `READY_ACK` messages, but this is operating outside of the framework and there's only so much we can do to keep that from occurring.

In React because `IO` can occur in the view, and `IO` can differ between backends there could be loss of synchronicity, on initial draw. But I think the handshake + `VTree` creation purity mitigates this risk.

PR #1586, opened 2026-08-17
fix: MTS bubbling incorrectly suppressed BTS delivery 

dmjio:
## Motivation

- `routeEvent` in `ts/miso/native/mts/context.ts` used a single `fired` flag to decide whether to forward an event to BTS: once *any* `OnStatic` (main-thread) handler matched anywhere on the propagation chain, `fired` latched `true` for the rest of the dispatch and the event was never forwarded to BTS — even when that handler had not requested `stopPropagation`.
- This broke ordinary DOM-style bubbling across the MTS/BTS boundary: a child with a non-stopping `OnStatic` handler would silently swallow the event before it ever reached a plain (background-thread) handler on an ancestor.
- Verified against ReactLynx's own dual-thread event dispatch (`dispatch_event_by_path` in `web-core`), which gates propagation per-node via an explicit "catch" handler, not via a single whole-dispatch "something matched" flag — confirming this was a bug, not an intentional design choice.

## Fix

- `mts/context.ts`: replaced the `fired` flag with `stopped`, only set when a matched handler's `options.stopPropagation` is `true`. Forwarding to BTS now happens whenever nothing stopped propagation, regardless of whether a non-stopping `OnStatic` handler fired along the way.
- `event.ts`: `onWithOptions` sets a live `runEvent` callback on every event object regardless of whether it's `OnStatic`-marked, so forwarding unconditionally would let BTS's own `delegateEvent` fire the *same* handler a second time. Guarded all three lookup sites (mid-chain VNode branch, target-node captures/bubbles branch, and `propagateWhileAble`) to skip `runEvent` whenever the event object carries a `staticKey` — that handler is main-thread-owned and was already run via `dispatchMainThreadEvent` — while still letting propagation continue past that node so ancestor handlers still fire.
- Updated `native-mts.spec.ts`'s `routeEvent` test, which had the old (buggy) behavior encoded as an assertion (`btsDispatches.length` expected `0`); it now expects the event to reach BTS.
- Regenerated `js/miso*.js` bundles from the updated TypeScript.

All 378 tests in `ts/spec` + `ts/miso.spec.ts` pass.


dmjio:
This incorrectly forwards all MTS events to BTS, closing.

PR #1598, opened 2026-08-22
`fix`: escape text nodes and attribute values in the `SSR` renderer (`ToHtml`)

dmjio:
Previously `Miso.Html.Render.toHtml` emitted `VText` content and attribute values verbatim. Any text containing '<' followed by an ASCII letter or '/' (code listings, user-generated content, haddock-style `\<a\>` in doc text, and so on) therefore parsed as real markup on the client:

  * the browser materializes phantom elements inside the prerendered page (e.g. a stray `<a>` element in the middle of a `<code>` block),
  * hydration detects "VTree differed from node", cannot copy the DOM into the virtual DOM, and falls back to a full diff -- defeating SSR,
  * any untrusted string rendered as a text node or attribute value is a markup / script injection (XSS) vector.

Observed in the wild in miso-ui: a rendered Haskell code sample containing `(@\<a\>@)` corrupted the prerendered kitchen sink and broke hydration (highlight.js also flagged the block with its "unescaped HTML" security warning).

The renderer now escapes:

  * text nodes: `&` `<` `>` become `&amp;` `&lt;` `&gt;`
  * attribute values (Property, ClassList, Styles): the same, plus `"` becomes `&quot;` (values are always rendered double-quoted)

Escaping is skipped inside raw-text elements (`script`, `style`), where the HTML parser does not decode character references and escaping would corrupt inline JS / CSS. The raw-text flag is threaded through `VFrag` and `VComp`/`VCompStatic` children via the new `renderBuilderWith`.

Because the browser decodes entities back to the original characters, the hydrated DOM text is now byte-for-byte identical to the client's virtual DOM -- consistent with the adjacent-text-node collapsing that toHtml already performs for hydration fidelity.

Internal cleanup: `toHtmlFromJSON` becomes `textFromJSON` (returning MisoString instead of Builder) so all attribute values flow through a single escaping point. It was not exported.

Verified with GHC 9.12.2 (VANILLA/native):

  toHtml (div_ [] [ "a <a> & <b>" ])
    ==> <div>a &lt;a&gt; &amp; &lt;b&gt;</div>
  toHtml (script_ [] "if (a < b && c > d) {}")
    ==> <script>if (a < b && c > d) {}</script>
  toHtml (div_ [ textProp "data-x" "say \"hi\" & <bye>" ] [])
    ==> <div data-x="say &quot;hi&quot; &amp; &lt;bye&gt;"></div>

Note: code that relied on smuggling raw HTML through text nodes will now see it escaped; that use case wants an explicit rawHtml-style escape hatch rather than being the default behavior of `text`.

### DIFF
--- a/src/Miso/Html/Render.hs
+++ b/src/Miso/Html/Render.hs
@@ -46,8 +46,16 @@
 -- * __'Miso.Types.VNode'__ — rendered as @\<tag attrs\>children\<\/tag\>@.
 --   Self-closing elements (@\<br\/\>@, @\<img\/\>@, @\<input\/\>@, …) are
 --   rendered without a closing tag.
--- * __'Miso.Types.VText'__ — rendered as a raw text string (no escaping
---   beyond what is already in the 'Miso.String.MisoString').
+-- * __'Miso.Types.VText'__ — rendered with HTML escaping: @&@, @\<@ and @>@
+--   become @&amp;@, @&lt;@ and @&gt;@, so the browser's parser reproduces
+--   the original string exactly. This keeps hydration faithful (the parsed
+--   DOM text matches the client's virtual DOM byte-for-byte) and prevents
+--   markup \/ script injection through text content. Text inside raw-text
+--   elements (@script@, @style@) is emitted unescaped, since the parser
+--   does not decode entities there.
+-- * __Attribute values__ — rendered with HTML escaping: @&@, @\"@, @\<@ and
+--   @>@ become entities, for the same round-tripping and injection-safety
+--   reasons as text nodes.
 -- * __'Miso.Types.VComp'__ — recursively renders the sub-component's view
 --   using its initial (or hydrated) model.
 -- * __'Miso.Types.VFrag'__ — renders all children inline, no wrapper tag.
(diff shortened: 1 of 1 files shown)


dmjio:
Closing in favour of a follow-up that keeps this diagnosis and the tag-based mechanism, but avoids double-escaping.

The problem this fixes is real, and escaping at serialisation is the right place for it. But miso already escapes — at *construction*, in `text`, behind the `ssr` flag (`src/Miso/Types.hs:989`, the only `#ifdef SSR` in that module):

```haskell
text :: MisoString -> View context model action
#ifdef SSR
text = VText Nothing . htmlEncode
#else
text = VText Nothing
#endif
```

So with `-fssr` the two stack. `text "a & b"` renders `a &amp;amp; b` and displays as `a &amp; b`. That is the configuration this PR targets: `haskell-miso.org` builds its prerenderer with `+ssr` (`cabal.project`, gated to non-wasm/non-js) precisely so `text` HTML-encodes.

Two other notes for the follow-up:

* `textRaw` becomes a no-op, since `VText (Maybe Key) MisoString` carries no raw marker. That matters because `script_` is built on it (`Miso/Html/Element.hs:661`), and escaping JS inside `<script>` would corrupt it. The `rawTextElements = ["script","style"]` check in this PR already handles that correctly, and does it better than a node marker would — it matches how the HTML parser actually treats raw-text elements. It also fixes `style_`, which uses `text` and therefore has its CSS escaped under `+ssr` today (`a > b` becomes `a &gt; b`).
* `-fssr` does not close the hole this PR describes. It only reaches `text`/`vtext`. `text_`, `textKey`, `textKey_` and the `IsString` instance (`Types.hs:921`, not CPP-guarded at all) all put the string into `VText` verbatim — so a bare `"a <b>"` literal is unescaped even with the flag on. That is the likely path for the miso-ui corruption.

The follow-up will be this PR plus one deletion:

1. escape text nodes and attribute values in `renderBuilder`, skipping `script`/`style` by tag (as here);
2. drop `htmlEncode` from `text`, so escaping happens exactly once, at serialisation, independent of the cabal flag.

The attribute half of this PR has no conflict at all — `htmlEncode` was never applied to `Property` / `ClassList` / `Styles` values, so nothing escapes them in any configuration today.

PR #1605, opened 2026-08-24
feat: Eagerly free short-lived `JSVal` handles in the WASM runtime

dmjio:
On the WASM backend every `JSVal` carries a `Weak#` with a C finalizer so the JavaScript value can be released once the handle is collected. The RTS evacuates every weak pointer at the start of every GC, dead or alive, so a program that allocates hundreds of short-lived handles per frame (as `buildVTree` does) makes each GC pause scale with the number of handles created since the last one. Profiling `miso-mario` with `-hT` showed ~14 MB of `DEAD_WEAK` / `C_FINALIZER_LIST` being copied at every collection, producing ~100 ms pauses, while the application's own live data was ~50 KB.

Add `Miso.DSL.freeJSVal` (`GHC.Wasm.Prim.freeJSVal` on WASM, a no-op elsewhere) and use it to release handles the runtime creates and nothing else can reach:

  * `(#)`, `call`, `new`: the looked-up function and the argument array.
  * `updateRef`, `populateClass`, `getDrawingContext`, `diff`, `flush`: the `jsg "miso"` handle and the ignored call result.
  * `createNode`: the `css`/`props`/`events`/`captures`/`bubbles` scratch objects once they are attached to the `vnode`.
  * `setAttrs`: `getProp` results and Property values freshly marshalled from `Number`/`Bool`/`Array`/`Object` (never `String`, which shares the caller's `JSString` handle, nor `Null`, which is a shared constant).
  * `buildVTree`: the children array, the node type, and child node handles once linked into the parent -- except nodes with event handlers (the handler closure reads the node at event time) and components (their callbacks close over the component object).
  * `onWithOptions`: its scratch handles; `decodeAtVal`, the callback and the node stay alive.

Handles received from the caller (strings in the `View`, in particular) are never freed. See Note [Freeing `VTree` handles] in `Miso.Runtime`.

Measured on `miso-mario` (`-H64m -A32m`): `C_FINALIZER_LIST` per GC 7.3 MB ->
2.3 MB. `DEAD_WEAK` is unchanged because `finalizeWeak#` cannot unlink from the singly-linked weak list; `markWeakPtrList` still evacuates dead weaks, which is an RTS limitation to be addressed upstream.

### DIFF
--- a/ffi/ghc/Miso/DSL/FFI.hs
+++ b/ffi/ghc/Miso/DSL/FFI.hs
@@ -105,6 +105,9 @@ isUndefined_ffi = undefined
 -----------------------------------------------------------------------------
 freeFunction_ffi :: JSVal -> IO ()
 freeFunction_ffi = undefined
+
+freeJSVal_ffi :: JSVal -> IO ()
+freeJSVal_ffi _ = pure ()
 -----------------------------------------------------------------------------
 -- | Schedules a callback to run before the next repaint.
 --
--- a/ffi/js/Miso/DSL/FFI.hs
+++ b/ffi/js/Miso/DSL/FFI.hs
@@ -66,6 +66,7 @@ module Miso.DSL.FFI
   , isNull_ffi
   , jsNull
   , freeFunction_ffi
+  , freeJSVal_ffi
   , listProps_ffi
   , requestAnimationFrame
   , cancelAnimationFrame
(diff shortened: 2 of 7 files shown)


dmjio:
image

## What the branch delivers

- Unbounded leak eliminated — exported event-handler callbacks are freed when their vtree is replaced (previously only reclaimed if JS's `FinalizationRegistry` ever noticed, which it didn't keep up with). Long-running apps no longer grow.
- ~30 % less copied per GC — every handle the runtime creates (`createNode` scratch objects, `#`/`call`/`new` temporaries, children arrays, property values, per-frame `diff`/`flush`/`updateRef` handles) is freed eagerly instead of surviving as a finalizer-list record.
- New public API: `Miso.DSL.freeJSVal` (no-op off-wasm), documented with the ownership rules.

PR #1634, opened 2026-09-04
fix: reused `Component` keeps a stale `view` closure after a parent rerender

ners:
## Problem

When a parent re-renders, the diff reuses an already-mounted child `VComp` (`ts/miso/dom.ts`, the `n.key === c.key` branch) so it keeps its `model`. But the rebuilt `VComp` carries a freshly built `Component`, and its `view` is discarded: `diffProps` only propagates `props`, and `initialize` baked the original `view` into `_componentDraw`.

So a child whose `view` closes over parent-side data renders what it first mounted with, forever. Nothing errors; the model updates, `dirtyCheck` reports dirty, and the component redraws _with the stale closure_.

This breaks `Miso.Fluent.translate`, which mounts a component per message:

```haskell
translate ref f = mountUseContext $ component () noop \ctx () () ->
    f $ translate ref (getBundle ctx)
```

## Repro

```haskell
main :: IO ()
main = startApp defaultEvents (component (0 :: Int) (\() -> this += 1) view')

view' :: (Eq ctx) => ctx -> props -> Int -> View ctx Int ()
view' _ _ n =
  div_ []
    [ button_ [onClick (), id_ "bump"] [text ("bumps: " <> ms n)]
    , div_ [id_ "plain"] [text ("plain: " <> ms n)]
    , div_ [id_ "nested"] [nested n]
    ]

nested :: (Eq ctx) => Int -> View ctx model action
nested n = mountUseContext . component () noop $ \_ () () -> text ("nested: " <> ms n)
```

Click `#bump`: `plain` counts up, `nested` stays at `nested: 0`.

## Fix

Store the `view` in an `IORef` on `ComponentState`, read it in `_componentDraw`, and have `diffProps` (which already runs on every reuse and holds the new `Component`) adopt it, redrawing when it changed. Identity is compared by `StableName`, so a `view` that is a shared top-level binding compares equal and nothing is redrawn.

Two things the redraw must avoid:

- `enqueueSchedule` runs the props phase, firing `onPropsChanged` for a component whose `props` did not change.
- the context-propagation sentinel redraws every `useContext` component, which rebuilds their children's closures and reflags them, never reaching a fixed point. Redraws are targeted instead, so the cascade terminates at the leaves.

The cascade also skips the per-component `requestAnimationFrame` wait; otherwise a page with one mounted component per translated message takes a frame each to settle.

### DIFF
--- a/src/Miso/Runtime.hs
+++ b/src/Miso/Runtime.hs
@@ -247,6 +247,7 @@ initialize events _componentParentId hydrate isRoot live initialProps maybeKey _
 
   _componentDOMRef <- getComponentMountPoint
   _componentVTree <- newIORef (VTree (Object jsNull))
+  _componentView <- newIORef view
   _componentSubThreads <- newIORef M.empty
 
   frame <- newEmptyMVar :: IO (MVar Double)
@@ -256,16 +257,20 @@ initialize events _componentParentId hydrate isRoot live initialProps maybeKey _
     asyncCallback1 $ \jsval -> do
       putMVar frame =<< fromJSValUnchecked jsval
 
-  let _componentDraw = \newModel -> do
+  let drawWith = \awaitFrame newModel -> do
         currentProps <- (^. componentProps) . (IM.! _componentId) <$> readIORef components
         currentContext <- readIORef globalContext
+        -- Not the @view@ bound above: a parent re-render can replace it.
+        currentView <- readIORef _componentView
         newVTree <-
           buildVTree events _componentParentId _componentId Draw live
-            _componentSink logLevel newModel (view currentContext currentProps newModel)
+            _componentSink logLevel newModel (currentView currentContext currentProps newModel)
         newHandlers <- collectEventHandlers
         oldVTree <- readIORef _componentVTree
-        _frame <- requestAnimationFrame rAFCallback
-        _timestamp :: Double <- takeMVar frame
+        when awaitFrame $ do
+          _frame <- requestAnimationFrame rAFCallback
+          _timestamp :: Double <- takeMVar frame
+          pure ()
         Diff.diff (Just oldVTree) (Just newVTree) _componentDOMRef
         FFI.updateRef oldVTree newVTree
         atomicWriteIORef _componentVTree newVTree
@@ -273,6 +278,8 @@ initialize events _componentParentId hydrate isRoot live initialProps maybeKey _
         -- See Note [Freeing event handler callbacks].
         swapEventHandlers _componentId newHandlers
         FFI.flush
+      _componentDraw = drawWith True
+      _componentRedraw = drawWith False
 
 #ifdef NATIVE
   -- N.B. all three cross-thread dispatch functions below wrap their FFI call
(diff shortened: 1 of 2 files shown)


dmjio:
I believe the `view` will rerender (not be stale) if we utilize the props feature, via `mountWithProps`. The issue above is that `n` is being captured from the parent state (this is a footgun if defining `Component` inline), so it exists as an immutable heap object from the child's perspective. The child (`nested` here) is guaranteed to re-render if the parent state is passed in as props to the child (via `mountWithProps`).

image

Sauce

```haskell
main :: IO ()
main = startApp defaultEvents top

top :: Component () () Int ()
top = component model_ update_ view_
  where
    model_ :: Int
    model_ = 0

    view_ _context _props n =
      vfrag
        [ button_
          [ onClick ()
          , id_ "bump"
          ]
          [ text ("bumps: " <> ms n)
          ]
        , div_
          [ id_ "plain"
          ]
          [ text ("plain: " <> ms n)
          ]
        , div_
          [ id_ "nested"
          ]
          [ mountWithProps n nested -- dmj: passes in 'n'
          ]
        ]

    update_ () =
      this += 1

nested :: Component context Int () action -- props ~ Int
nested = component model_ update_ view_
  where
    model_ =
      ()
    update_ =
      noop
    view_ _context props _model = -- dmj: 'n' above is now bound to 'props'
      text ("nested: " <> ms props)
```

dmjio:
Closing as solved w/ `mountWithProps` and tested locally. Please re-open if the issue still persists.

ners:
`mountWithProps` works in your example because `n` is `Eq` and `nested` is a static top-level `Component`. Its `view` never changes, only `props` does.

My call sites can't take that path. Both `Miso.Fluent.translate` and our own `withContext mount` via `mountUseContext`, which hardcodes `props ~ ()`:
```haskell
mountUseContext :: (Eq context, Eq childModel) => Component context () childModel childAction -> View context model action
```

There's no `props` var to route data through, and the mounted `view` isn't static.
It closes over an arbitrary caller-supplied function.

Minimal repro:
```haskell
withContext :: (Eq context) => (context -> View context () action) -> View context model action
withContext f = mountUseContext $ component () noop $ \ctx () () -> f ctx

main :: IO ()
main = startApp defaultEvents (component (0 :: Int) (\() -> this += 1) view')

view' :: (Eq ctx) => ctx -> props -> Int -> View ctx Int ()
view' _ _ n =
  div_ []
    [ button_ [onClick (), id_ "bump"] [text ("bumps: " <> ms n)]
    , div_ [id_ "nested"] [withContext (\_ctx -> text ("nested: " <> ms n))]
    ]
```

Click `#bump`: `nested` still reads `0`. There's no `mountWithProps` call to make here. `withContext`'s signature has no `props` parameter, and adding one pushes the problem onto every call site instead of fixing it once. `Reference` (what `translate` closes over) isn't `Eq` either, so the same applies there.

dmjio:
The context isn't actually getting modified in this case. The parent `model` is, but that is component local state. The parent would need to call `modifyContext_ (this += 1)` in this case. Then the child would rerender. Otherwise passing the parent model state to the child is only valid at mount time. Rerendering in response to parent model changes is only supported by props and context explicitly.

dmjio:
Also the child component won't automatically unmount and remount unless the key is changing in response to every parent model change. In this case no key is specified so the component won't unmount unless it's replaced.

